Issue #251, opened 2019-11-22
Supervisor view for study requests


candu:
**Description**
Right now, we only show requests submitted by the current user.  #249 will add full CRUD functionality to that view.

However, for the supervisor side of this, that's not enough: they need to be able to CRUD requests submitted by _all_ users.  We don't currently have ADFS authentication, but in the meantime we can create this view, and "hide" it behind a separate URL that isn't made accessible from the interface.

**Acceptance Criteria**
- [x] add `submitted by` column to Track Requests table;
- [x] create new route for supervisor view;
- [x] in supervisor view, fetch _all_ requests from backend for table generation.

**Additional Notes**
We can probably get away without any kind of access control / gating on the relevant endpoints for now (e.g. unguessable random URLs, magic tokens, passwords, etc.)  This will of course change once we have role-based access control in place, but there's not much point in hand-rolling our own hacky RBAC with ADFS so close on the horizon.

Issue #250, opened 2019-11-22
Search functionality for Track Requests


candu:
**Description**
The Track Requests table is currently sortable, but users cannot yet search for specific values of specific columns.  This will become especially important once we implement the supervisor view.

This task adds per-column search boxes that can be used to filter on specific values for each column.

**Acceptance Criteria**
- [x] add `searchable` option on column definitions in `FcCardTable`, and show search boxes for `searchable` columns;
- [x] wire those search boxes up to filter state, and update that state when user interacts with search boxes;
- [x] use filter state to filter results in table.

**Additional Notes**
For now, we can ignore pagination and simply fetch all available requests.

Eventually we'll want full database-backed pagination, but we can actually get a fair distance without it.  An intermediate state is to still fetch everything in the backend, but implement the pagination interface in the frontend (e.g. using `.slice()` on the returned list).

Backend pagination probably implies a larger rethink of how `FcCardTable` works, or finding some kind of pagination component that we can pull in.

PR #254, opened 2019-11-25
Add Requested By to request table

candu:
This PR starts work on #250 and #251 by introducing a "Requested By" column to the list of requests in Track Requests.

To make this work, I've added `UserController` and `UserDAO.bySubjects`, which perform multi-lookups of users by their `subject` field.  This is then used by the `itemsStudyRequests` getter to populate `requestedBy`, which is in turn used by `FcCardTableRequests`.

### DIFF
--- a/lib/db/UserDAO.js
+++ b/lib/db/UserDAO.js
@@ -19,12 +19,7 @@ class UserDAO {
     const uniqueSubjects = Array.from(new Set(subjects));
     const sql = 'SELECT * FROM "users" WHERE "subject" IN ($(uniqueSubjects:csv))';
     const rows = await db.manyOrNone(sql, { uniqueSubjects });
-    const users = new Map();
-    rows.forEach((user) => {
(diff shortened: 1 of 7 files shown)


lib/db/UserDAO.js
```diff
@@ -1,11 +1,32 @@
...
+   * Map user subjects to users.
+   *
+   * @param {Array<string>} subjects - array of user subjects
+   * @returns {Promise<Map<string, Object>>} map from user subjects to users
+   */
+  static async bySubjects(subjects) {
+    const uniqueSubjects = Array.from(new Set(subjects));
+    const sql = 'SELECT * FROM "users" WHERE "subject" IN ($(uniqueSubjects:csv))';
+    const rows = await db.manyOrNone(sql, { uniqueSubjects });
+    const users = new Map();
```
candu: `return new Map(rows.map(user => [user.subject, user]));`

PR #255, opened 2019-11-26
Column-level search bars in FcCardTable

### LINKED ISSUE #250: Search functionality for Track Requests

candu:
This PR closes #250 by adding column-level search functionality to `FcCardTable`:

search_table

It does not attempt to handle the interaction between this and selection - that will likely be handled as part of the work on supervisor view.


lib/Constants.js
```diff
@@ -366,9 +366,27 @@ const RequestStatus = {
 const SearchKeys = {
   Requests: {
     ID: (q, r) => r.id.toString() === q,
-    LOCATION: (q, r) => r.location.description.indexOf(q) !== -1,
-    REQUESTER: (q, r) => r.requestedBy.name.indexOf(q) !== 1,
-    STATUS: (q, r) => r.status.indexOf(q) !== 1,
+    LOCATION: (q, r) => {
+      if (r.location === null) {
```
candu: we should really guard against falsy values for `r.location.description` here.

lib/Constants.js
```diff
@@ -366,9 +366,27 @@ const RequestStatus = {
...
+        return q === '';
+      }
+      const qLower = q.toLowerCase();
+      const rLower = r.location.description.toLowerCase();
+      return rLower.indexOf(qLower) !== -1;
+    },
+    REQUESTER: (q, r) => {
+      if (r.requestedBy === null) {
```
candu: same here: guard against falsy `r.requestedBy.name`.  in addition, we should change the checks for `r.location` and `r.requestedBy` to catch all falsy values (`null`, `undefined`, empty string).

PR #256, opened 2019-11-26
Supervisor view for Track Requests

### LINKED ISSUE #251: Supervisor view for study requests

candu:
This PR closes #251 by adding a query parameter `?isSupervisor=true` throughout the study request track, view, and edit pages.

If present with _any_ value, this query param:

- sends `?isSupervisor=true` to relevant REST API CRUD endpoints;
- ensures that all links that keep the user within those pages preserve the `isSupervisor` flag.

### DIFF
--- a/lib/db/StudyRequestDAO.js
+++ b/lib/db/StudyRequestDAO.js
@@ -79,6 +79,43 @@ SELECT
   }
 
   /**
+   * Fetch all study requests.
+   *
+   * @returns {Array<Object>} array of all study request objects
+   */
+  static async all() {
+    const sql = `
+SELECT
+  "id",
+  "createdAt",
+  "userSubject",
+  "status",
+  "serviceRequestId",
+  "priority",
+  "dueDate",
+  "estimatedDeliveryDate",
+  "reasons",
+  "ccEmails",
+  "centrelineId",
+  "centrelineType",
+  ST_AsGeoJSON("geom")::json AS "geom"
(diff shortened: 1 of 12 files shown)


lib/db/StudyRequestDAO.js
```diff
@@ -78,6 +78,43 @@ SELECT
     });
   }
 
+  /**
+   * Fetch all study requests.
+   *
+   * @returns {Array<Object>} array of all study request objects
+   */
+  static async all() {
```
candu: we should eventually add tests for `StudyRequestDAO.all`, `StudyRequestDAO.byUser` here.

PR #267, opened 2019-12-03
Status and assignment updates for study requests

### LINKED ISSUE #266: Status and assignment for supervisor view
**Description**
Supervisors should be able to set request status and assign it to people.

**Acceptance Criteria**
Basic status updates:
- [x] add "approve" action in requests table, which sets request status to "Accepted";
- [x] add "reject" action in requests table, which sets request status to "Rejected";
- [x] add "complete" action in requests table, which sets request status to "Completed";

Assignment:
- [x] database schema is changed to include `"assignedTo" VARCHAR` column, and to have a separate `study_request_assignees` table that contains OTI and Field Staff;
- [x] requests table includes "Assigned To" column, which defaults to "Not Assigned" (`NULL` value in database);
- [x] in supervisor mode, "Assigned To" is a dropdown with OTI, Field Staff, and Not Assigned as options;
- [x] when supervisor selects a value other than "Not Assigned", set request status to "In Progress";

Permissions:
- [x] frontend hides status / assignment update features for non-supervisors, but shows changes in status / assignment;
- [x] backend disallows status / assignment updates for non-supervisors.

**Additional Notes**
Don't worry about email notification for now; just update the status.

candu:
This PR closes #266 by adding assignment to the study request data model, and by providing controls to supervisors for updating status and assignment.

In the process, we make a couple of small usability improvements to disabled button states and dropdowns.


.eslintrc.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     'arrow-parens': ['error', 'as-needed', { requireForBlockBody: true }],
     'max-classes-per-file': 'off',
     'no-console': process.env.NODE_ENV === 'production' ? 'error' : 'off',
-    'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
+    'no-debugger': 'error',
```
candu: this is largely due to some errant `debugger` statements found in `DAO.spec.js`.

we left `no-console` as-is, as making it always `'error'` would break dev builds using `console.log()`.  our "CI-ish" deployment script catches broken AWS-mode builds.

lib/Constants.js
```diff
@@ -334,38 +334,34 @@ const RequestStatus = {
...
   },
-  CANCELLED: {
-    icon: 'times-circle',
-    variant: 'error',
-  },
 };
 
 const SearchKeys = {
   Requests: {
+    ASSIGNED_TO: (q, r) => {
```
candu: please add test cases in `Constants.spec.js` to maintain coverage

lib/controller/StudyRequestController.js
```diff
@@ -176,6 +176,12 @@ StudyRequestController.push({
     if (studyRequestNew.userSubject !== studyRequestOld.userSubject) {
       return Boom.badRequest('cannot change owner for study request');
     }
+    if (studyRequestNew.assignedTo !== studyRequestOld.assignedTo && !isSupervisor) {
```
candu: use `Boom.forbidden()` for these use cases - at this point, these are semantically valid requests, it's just that the user doesn't have permission to do them.

scripts/db/schema-5.down.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+
+ALTER TABLE "study_requests" DROP COLUMN "assignedTo";
+
+DROP TABLE "study_request_assignees";
+
+DELETE FROM "study_request_reasons" WHERE "value" = 'AWS';
```
candu: can we get rid of this `AWS` reason?  these should probably be rethought, and randomly adding individual reasons isn't a good way to do that.  also, this doesn't relate to the larger change here in any meaningful way.

scripts/db/schema-5.up.sql
```diff
@@ -0,0 +1,21 @@
+BEGIN;
+
+INSERT INTO "study_request_reasons"
```
candu: same here: get rid of this

tests/jest/db/DAO.spec.js
```diff
@@ -426,6 +424,7 @@ test('StudyRequestDAO', async () => {
     status: 'REQUESTED',
     serviceRequestId: null,
     priority: 'STANDARD',
+    assignedTo: null,
```
candu: please run this test to ensure that it still works (it's not part of pre-commit checks, since it's quite slow)

web/components/FcModalRequestStudyConfirmation.vue
```diff
@@ -5,24 +5,35 @@
     </template>
     <template v-slot:content>
       <p v-if="update">
-        Your request has been updated.
+        <span v-if="isSupervisor">
```
candu: can we test the non-supervisor update case manually as well?

PR #268, opened 2019-12-05
Dynamically generated vector tiles: counts, collisions, and schools

### LINKED ISSUE #265: Investigate alternative approaches for vector tile generation
**Description**
In [Vector tile spike](https://www.notion.so/Vector-tile-spike-1357ad906c26421fb66f98c33bccd0fc), @candu explored `ST_AsMVT` as an approach for dynamic tile generation.  That turned out to not be viable due to the tangle of dependencies around `libprotobuf`, and due to the fact that AWS doesn't support `ST_AsMVT` until PostgreSQL 10.5 - meaning a major version upgrade would need to be coordinated with Cloud Services, and we already have enough blocker points there.

There's a possible alternative, however!  `ST_AsMVTGeom` shouldn't require `libprotobuf`, so we can use it to clip the geometry; we can then either:

- build the vector tiles in the web application server; or
- pass geometry down to the client and build vector tiles there!

If even that doesn't work, we might be able to find a GeoJSON-based approach as a fallback.

**Acceptance Criteria**
- [ ] assess other approaches for vector tile generation;
- [ ] if possible, use a working approach to switch counts over to a vector-tile-based approach.

**Additional Notes**
Right now, our counts layer is loaded via a bounding-box REST API endpoint.  This provides an opportunity to get the quick win of more efficient loading there, while also exploring viability here.

candu:
This PR closes #265 by successfully implementing dynamic generation of vector tiles.

This implementation is then used to build vector tiles for counts, collisions, and schools, as a proof-of-concept of this process.

The core of the implementation consists of two key pieces.  First, a general format for SQL queries that fetches relevant features for a tile, then transforms feature geometry into tile coordinate space:

```sql
WITH features AS (
  SELECT
    geom,
    -- other columns
  WHERE ST_Intersects(
    ST_Transform(geom, 3857),
    ST_MakeEnvelope($(xmin), $(ymin), $(xmax), $(ymax), 3857)
  )
)
SELECT ST_AsGeoJSON(
  ST_Intersection(
    ST_SnapToGrid(
      ST_Affine(
        ST_Simplify(
          ST_RemoveRepeatedPoints(ST_Transform(geom, 3857), $(res)),
          $(res),
          FALSE
        ),
        $(fx), 0,
        0, $(fy),
        $(xoff), $(yoff)
      ),
      0, 0, 1, 1
    ),
    ST_MakeEnvelope($(bmin), $(bmin), $(bmax), $(bmax), 3857)
  )
)::json AS geom,
-- other columns
FROM features;
```

Second, `lib/geo/VectorTile` provides a way to convert feature geometries and attributes from PostgreSQL to a form compatible with `vt-pbf`, which then writes out the vector tile protobuf.

### DIFF
--- a/lib/controller/DynamicTileController.js
+++ b/lib/controller/DynamicTileController.js
@@ -1,0 +1,55 @@
+import Joi from '@hapi/joi';
+import vtpbf from 'vt-pbf';
+
+import DynamicTileDAO from '@/lib/db/DynamicTileDAO';
+import VectorTile from '@/lib/geo/VectorTile';
+
+/**
+ * Endpoints for dynamic tile generation.
+ *
+ * @type {Array<Hapi.ServerRoute>}
+ */
+const DynamicTileController = [];
+
+/**
+ * Get the given tile.
+ *
+ * @memberof DynamicTileController
+ * @name getTestDynamicTile
+ * @type {Hapi.ServerRoute}
+ */
+DynamicTileController.push({
+  method: 'GET',
+  path: '/dynamicTiles/{layerName}/{z}/{x}/{y}.pbf',
+  options: {
+    validate: {
+      params: {
+        layerName: Joi.string().required(),
+        z: Joi.number().integer().min(0).required(),
+        x: Joi.number().integer().min(0).required(),
+        y: Joi.number().integer().min(0).required(),
+      },
+    },
+  },
+  handler: async (request, h) => {
+    const {
+      layerName,
+      z,
+      x,
+      y,
+    } = request.params;
+    const features = await DynamicTileDAO.getTileFeatures(layerName, z, x, y);
+    const tile = new VectorTile(layerName, features);
+    const tileData = vtpbf(tile);
+    const tileBuffer = Buffer.from(tileData);
+    /*
+     * See https://github.com/mapbox/vector-tile-spec/tree/master/2.1#22-multipurpose-internet-mail-extensions-mime
+     * for preferred MIME type here.
+     */
+    return h.response(tileBuffer)
+      .encoding('binary')
+      .type('application/vnd.mapbox-vector-tile');
+  },
+});
+
+export default DynamicTileController;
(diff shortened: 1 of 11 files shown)


lib/controller/DynamicTileController.js
```diff
@@ -0,0 +1,56 @@
...
+ * @type {Hapi.ServerRoute}
+ */
+DynamicTileController.push({
+  method: 'GET',
+  path: '/dynamicTiles/{layerName}/{z}/{x}/{y}.pbf',
+  options: {
+    auth: { mode: 'try' },
+    validate: {
+      params: {
+        layerName: Joi.string().required(),
```
candu: provide a `.valid()` clause on this

lib/db/DynamicTileDAO.js
```diff
@@ -0,0 +1,180 @@
...
+    if (layerName === 'collisions') {
+      return DynamicTileDAO.getCollisionsFeatures(tileInfo);
+    }
+    if (layerName === 'counts') {
+      return DynamicTileDAO.getCountsFeatures(tileInfo);
+    }
+    if (layerName === 'schools') {
+      return DynamicTileDAO.getSchoolsFeatures(tileInfo);
+    }
+    // TODO: throw an error here
```
candu: yes, please do.

web/components/PaneMap.vue
```diff
@@ -204,41 +188,44 @@ function injectSourcesAndLayers(rawStyle, dataCountsVisible) {
...
     type: 'circle',
-    source: 'counts-visible',
-    filter: ['!', ['has', 'point_count']],
+    minzoom: ZOOM_TORONTO,
+    maxzoom: ZOOM_MAX + 1,
     paint: {
-      'circle-color': PAINT_COLOR_COUNTS,
-      'circle-opacity': PAINT_OPACITY,
-      'circle-radius': PAINT_SIZE_COUNT_POINTS,
+      'circle-color': '#777777',
```
candu: let's change this to a more visible colour for now (later, we'll probably want to have an icon of sorts)

Issue #269, opened 2019-12-05
Layer tweaks: collisions, counts, schools


candu:
**Description**
Now that we have a basic end-to-end implementation of vector tiles for collisions, counts, and schools, we need to make some adjustments to get these ready for deployment.

**Acceptance Criteria**
- [x] counts are rolled up into clusters, as before;
- [x] collisions are marked as KSI vs. non-KSI;
- [x] collision heatmap tiles are pre-generated, to reduce load on database;
- [x] schools are displayed with a school icon.

**Additional Notes**
These are intended to be relatively quick changes, mainly so that we can reduce load on the webserver from dynamic tile generation.

PR #270, opened 2019-12-09
DAG for collisions heatmap tile generation

candu:
This PR makes progress on #269 by introducing the `collisions_vector_tiles` DAG in Airflow, which builds vector tiles covering the last year of collisions Toronto-wide for zoom levels 10-15.

These tiles contain a property `heatmap_weight` that is summed during `tippecanoe`'s clustering.

### DIFF
--- a/scripts/airflow/tasks/build_collisions_tiles/download_collisions.sql
+++ b/scripts/airflow/tasks/build_collisions_tiles/download_collisions.sql
@@ -1,0 +1,36 @@
+COPY (
+WITH event_injury AS (
+  SELECT i.collision_id, MAX(CONCAT('0', i.injury)::int) AS injury
+  FROM collisions.events e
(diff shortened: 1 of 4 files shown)


scripts/airflow/tasks/build_collisions_tiles/download_collisions.sql
```diff
@@ -0,0 +1,36 @@
+COPY (
+WITH event_injury AS (
+  SELECT i.collision_id, MAX(CONCAT('0', i.injury)::int) AS injury
+  FROM collisions.events e
+  JOIN collisions.involved i ON e.collision_id = i.collision_id
+  WHERE e.accdate >= now() - interval '1 year'
```
candu: we might eventually need to make this configurable, but for now this is OK

PR #271, opened 2019-12-09
Layer tweaks for collisions, counts, and schools

### LINKED ISSUE #269: Layer tweaks: collisions, counts, schools

candu:
This PR closes #269 by making several small performance and visualization changes to the collisions, counts, and schools layers.

It adds a collisions heatmap visible at lower zoom levels (10-15), with heavy weighting on KSI collisions.  At higher zoom levels, this heatmap fades into individual points, with KSI collisions shown larger in a deeper red.

The school layer is now rendered using Font Awesome icons - this gives us a wide range of icons to play with in prototyping layers.  (We may eventually want to switch these out for nicer icons.)

Counts are rolled up by location to reduce the response size from `/api/dynamicTiles/counts`.

In addition, this PR tests out the proposed switch from at-marker popups to an info box at top-left.  The idea is that this will reduce clutter and make mouse gestures easier, and that it prepares us for future updates to the interface.

### DIFF
--- a/scripts/deployment/etl/nginx/default.d/glyphs.conf
+++ b/scripts/deployment/etl/nginx/default.d/glyphs.conf
@@ -1,0 +1,5 @@
+location /glyphs {
+    alias /data/glyphs;
+    autoindex on;
+    add_header Access-Control-Allow-Origin * always;
+}
--- a/bdit-flashcrow.code-workspace
+++ b/bdit-flashcrow.code-workspace
@@ -6,7 +6,7 @@
 			"jcbuisson.vue",
 			"ms-python.python",
 			"ms-vscode-remote.vscode-remote-extensionpack",
-			"shinnn.stylelint"
+			"thibaudcolas.stylelint"
 		]
 	},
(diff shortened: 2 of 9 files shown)


scripts/deployment/etl/nginx/default.d/glyphs.conf
```diff
@@ -0,0 +1,5 @@
+location /glyphs {
```
candu: we should eventually include a script for generating these glyphs...

Issue #284, opened 2019-12-16
Comments for study requests


candu:
**Description**
To help TSU supervisors and Traffic Ops investigators communicate around submitted requests, the most recent designs add commenting features to requests.

To support these changes, we'll need a `study_request_comments` table.

**Acceptance Criteria**
- [x] add database migration scripts for `study_request_comments`;
- [x] add relevant methods / DAOs and corresponding tests;
- [ ] add comments section under rest of study request details with dummy content;
- [ ] implement character counter logic for comment box;
- [x] implement necessary API endpoints, actions, etc. for submitting comments;
- [ ] implement comment deletion (author only).

**Additional Notes**
More details in [Figma](https://www.figma.com/file/hX4Dl2ieppwNf4btdAcjV3/Supervisor-View---Comments?node-id=16%3A0) and [Notion](https://www.notion.so/bditto/c887e7d20e2c4ce6bf0c55f2d59a6153?v=94b2df08e667426bb9b7d86751cf7e31&p=e476b1aa5f3b4f43b4a1c9c33c8fd837).

For `study_request_comments`, we could provide a separate DAO, or we could integrate this into `StudyRequestDAO`.  (It's probably wiser to have this as a separate DAO?  We can then fetch comments separately from the study request itself.)

Issue #283, opened 2019-12-16
Navigation and info hierarchy for Track Requests


candu:
**Description**
The current version of Track Requests needs a bit of copy / layout / interaction tweaking to match most recent designs.

To support these changes, we'll need to add an `closed` flag to `study_requests` in database.

**Acceptance Criteria**
- [x] add database migration scripts for `study_requests.closed`, and mark all existing records as not closed;
- [x] update `StudyRequestDAO` and associated tests to handle `closed` flag during creation
- [x] add actions / mutations in Vuex store as needed;
- [x] separate requests out into Open / Closed tabs on Track Requests page;
- [x] in Open / Closed tabs, allow requests to be closed / opened;
- [x] reorganize action buttons as shown in mocks.

**Additional Notes**
See [Notion](https://www.notion.so/bditto/c887e7d20e2c4ce6bf0c55f2d59a6153?v=94b2df08e667426bb9b7d86751cf7e31&p=732126c82b2e42bcb2f8d74a6b72d474) and [Figma](https://www.figma.com/file/VNPUJY3QMljIpqH1MvT2R6/Supervisor-View---Track-Requests?node-id=0%3A1) for more details.

Vue Material has the [Tabs](https://vuematerial.io/components/tabs) component, which also offers syncing with Vue Router.  From Vue Material docs, it is possible to [load individual components](https://vuematerial.io/getting-started/) without bringing in the whole library at once.  (This would be an OK place to start exploring Vue Material.)

PR #285, opened 2019-12-16
CRUD methods and API endpoints for comments

candu:
This PR makes progress on #283 and #284 by making the database, DAO layer, and REST API changes necessary to:

- handle opening / closing requests;
- manage comments on requests.

It also updates `DAO.spec.js` to cover these use cases.

### DIFF
--- a/lib/controller/StudyRequestController.js
+++ b/lib/controller/StudyRequestController.js
@@ -2,18 +2,22 @@ import Boom from '@hapi/boom';
 import Joi from '@hapi/joi';
 
 import StudyRequestDAO from '@/lib/db/StudyRequestDAO';
+import StudyRequestCommentDAO from '@/lib/db/StudyRequestCommentDAO';
 import EmailStudyRequestConfirmation from '@/lib/email/EmailStudyRequestConfirmation';
 import Mailer from '@/lib/email/Mailer';
 import LogTag from '@/lib/log/LogTag';
 import StudyRequest from '@/lib/model/StudyRequest';
+import StudyRequestComment from '@/lib/model/StudyRequestComment';
 import SuccessResponse from '@/lib/model/SuccessResponse';
 
 /**
- * CRUD handling for study requests.
+ * CRUD handling for study requests and related comments.
  *
  * @type {Array<Hapi.ServerRoute>}
  */
 const StudyRequestController = [];
+
+// STUDY REQUESTS
 
 /**
  * Create a new study request.  This endpoint also sends an email to the requester
@@ -34,7 +38,7 @@ StudyRequestController.push({
       schema: StudyRequest.read,
     },
     validate: {
-      payload: StudyRequest.transient,
+      payload: StudyRequest.create,
     },
   },
   handler: async (request) => {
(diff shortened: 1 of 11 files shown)


lib/controller/StudyRequestController.js
```diff
@@ -34,7 +38,7 @@ StudyRequestController.push({
       schema: StudyRequest.read,
     },
     validate: {
-      payload: StudyRequest.transient,
+      payload: StudyRequest.create,
```
candu: hmmm.  this probably means it wasn't validating beforehand, which means that we should double-check that our study request REST API endpoints still work after this change.

(we can test that in one sweep later, though.)

lib/controller/StudyRequestController.js
```diff
@@ -231,4 +236,214 @@ StudyRequestController.push({
...
+    const { studyRequestId } = request.params;
+    const { isSupervisor } = request.query;
+    const studyRequest = await StudyRequestDAO.byId(studyRequestId);
+    if (studyRequest === null) {
+      return Boom.notFound(`no study request found with ID ${studyRequestId}`);
+    }
+    if (studyRequest.userSubject !== subject && !isSupervisor) {
+      return Boom.forbidden('cannot comment on study request owned by another user');
+    }
+    const studyRequestComment = await StudyRequestCommentDAO.create(studyRequest, {
```
candu: just `return` this value

lib/controller/StudyRequestController.js
```diff
@@ -231,4 +236,214 @@ StudyRequestController.push({
...
+      return Boom.notFound(`no study request found with ID ${studyRequestId}`);
+    }
+    const commentOld = await StudyRequestCommentDAO.byId(commentId);
+    if (commentOld.studyRequestId !== studyRequest.id) {
+      return Boom.badRequest(
+        `no comment with ID ${commentId} for study request ${studyRequestId}`,
+      );
+    }
+
+    if (commentOld.id !== commentNew.id) {
```
candu: it feels kind of weird to do all this work to essentially update one field, but maybe this is OK if we ever need to add more fields to `study_request_comments`?

lib/controller/StudyRequestController.js
```diff
@@ -231,4 +236,214 @@ StudyRequestController.push({
...
+  handler: async (request) => {
+    const { subject } = request.auth.credentials;
+    const { studyRequestId } = request.params;
+    const { isSupervisor } = request.query;
+    const studyRequest = await StudyRequestDAO.byId(studyRequestId);
+    if (studyRequest === null) {
+      return Boom.notFound(`no study request found with ID ${studyRequestId}`);
+    }
+    if (studyRequest.userSubject !== subject && !isSupervisor) {
+      return Boom.forbidden('cannot comment on study request owned by another user');
```
candu: IMHO this is a sign that we should rethink the permissions / ownership model around study requests, studies, and such.

it probably makes sense for *all* users to be able to view and comment on study requests.  this allows for out-of-reporting-chain conversations (e.g. "hey, just so you're aware, we're requesting this study over here") and allows users to understand what other requests were made previously at a given location.

so here's a possible breakdown:

- study requests:
  - create: all staff;
  - read: all staff;
  - update: only supervisors or the owner, but:
    - only supervisors can approve, assign, change status;
  - delete: not possible
- studies:
  - create: only supervisors or the owner of the study request;
  - read: all staff;
  - update: only supervisors or the owner of the study request;
  - delete: only supervisors or the owner of the study request;
- comments:
  - create: anyone;
  - read: anyone;
  - update: only the author;
  - delete: only the author.

one issue here is that these are _authorization_ decisions, which are handled at the API layer; right now, however, we test only at the DAO layer.  to test this logic, we'll need to change that.

lib/controller/StudyRequestController.js
```diff
@@ -231,4 +236,214 @@ StudyRequestController.push({
...
+  handler: async (request) => {
+    const { subject } = request.auth.credentials;
+    const { studyRequestId } = request.params;
+    const { isSupervisor } = request.query;
+    const studyRequest = await StudyRequestDAO.byId(studyRequestId);
+    if (studyRequest === null) {
+      return Boom.notFound(`no study request found with ID ${studyRequestId}`);
+    }
+    if (studyRequest.userSubject !== subject && !isSupervisor) {
+      return Boom.forbidden('cannot view comments for study request owned by another user');
```
candu: see above: should be able to do this

lib/controller/StudyRequestController.js
```diff
@@ -231,4 +236,214 @@ StudyRequestController.push({
...
+      return Boom.badRequest('cannot change creation timestamp for study request');
+    }
+    if (commentOld.userSubject !== commentNew.userSubject) {
+      return Boom.badRequest('cannot change owner for study request');
+    }
+    if (commentOld.studyRequestId !== commentNew.studyRequestId) {
+      return Boom.badRequest('cannot change owner for study request');
+    }
+
+    if (commentOld.userSubject !== user.subject && !isSupervisor) {
```
candu: see above: only the author should be able to do this (not the supervisor!)

lib/controller/StudyRequestController.js
```diff
@@ -231,4 +236,214 @@ StudyRequestController.push({
...
+      return Boom.notFound(`no study request found with ID ${studyRequestId}`);
+    }
+    const comment = await StudyRequestCommentDAO.byId(commentId);
+    if (comment.studyRequestId !== studyRequest.id) {
+      return Boom.badRequest(
+        `no comment with ID ${commentId} for study request ${studyRequestId}`,
+      );
+    }
+
+    if (comment.userSubject !== user.subject && !isSupervisor) {
```
candu: see above: only the author should be able to do this (not the supervisor!)

Issue #293, opened 2020-01-09
Update auth flow and redirect URIs for QA ADFS application


candu:
**What Happened**
Our redirect URI for this application is configured to https://localhost:8080/api/auth/adfs-callback.  For testing purposes, it would help to add an entry on port 8081, as well as one for our dev and QA environments on AWS.

**What Should Happen**
The redirect URIs for the QA ADFS application should be:

- https://localhost:8080/api/auth/adfs-callback
- https://localhost:8081/api/auth/adfs-callback
- https://move.intra.dev-toronto.ca/api/auth/adfs-callback
- TODO: QA URL

`https://localhost:8081` facilitates usage in REST API tests, which spin up a standalone version of the MOVE application backend on port 8081.

**Acceptance Criteria**
- [x] Directory Services to reconfigure MOVE for authorization code flow;
- [x] MOVE to test authorization code flow in local development;
- [x] Directory Services to set redirect URIs for QA ADFS application as described above.
- [ ] MOVE to test redirect URI updates in local development, AWS dev, and AWS QA.

**Additional Notes**
When MOVE originally made this request, we'd envisioned using the `response_type=code` flow described [here](https://medium.com/@darutk/diagrams-of-all-the-openid-connect-flows-6968e3990660).  In this flow, the "client" is often a web application backend that handles communication with the "server", or authentication provider.

However, our ADFS application is currently configured to use the `response_type=id_token` flow.  This will need to be changed.

PR #294, opened 2020-01-09
ADFS integration [QA]

### LINKED ISSUE #162: Authenticate City of Toronto users [QA]
**Description**
We currently rely on a stub authentication method.  That's been sufficient for our prototypes, but changing that to real authentication is a prerequisite of several upcoming features:

- ability to track requests end-to-end, and to tie them to specific users;
- adding access controls (e.g. around MVA / PII data, or to view the entire request queue on the TSU side);
- bulk email to all users.

This task involves working with Cloud Services to identify our best path to authenticating City of Toronto users from EC2, then implementing that path in MOVE.

**Acceptance Criteria**
- [x] identify authentication method in consultation with Cloud Services;
- [x] work with Cloud Services to implement that method in MOVE;
- [x] strip out any unused authentication methods;

**Additional Notes**
We've previously discussed two options with Cloud Services: SAML and OpenID Connect.  We had previously waited for OpenID Connect to be ready, but at time of last conversation that was still in security review by an external vendor.  Any conversation with Cloud Services should start by checking on the progress of that process.

In the near future, we should pin down a "must-have-by" date on our side, to help meet our end-of-year goal to see early adopters incorporate MOVE into their workflow.  Initial estimate here would be sometime mid-late October, as that would leave enough time to plan application-specific auth work if necessary.

candu:
This PR closes #162 by providing a working end-to-end implementation of ADFS authentication in MOVE.

It includes changes to the `users` schema, as well as to study request-related tables that refer to `users`.  Since we're not yet putting this in front of people longer-term, and since the `sub` UUIDs previously generated won't line up with ADFS `sub` claim values, we simply drop and remake these tables.

Note that this depends on #293 to be fully functional.  Until that change is made, log in will not work, as the ADFS auth endpoint relies on a whitelist of callback URIs.  We have, however, tested it with the following flow:

- temporarily change the callback URI in `MoveConfig` to `/api/auth/adfs-callback`;
- when it redirects to `/api/auth/adfs-callback`, delete the leading `/api` from the path and load that page.

This should complete the ADFS flow as intended.

### DIFF
--- a/lib/db/StudyDAO.js
+++ b/lib/db/StudyDAO.js
@@ -53,7 +53,7 @@ SELECT
     const sql = `
 INSERT INTO "studies" (
   "createdAt",
-  "userSubject",
+  "userId",
   "studyRequestId",
   "studyType",
   "daysOfWeek",
@@ -62,7 +62,7 @@ INSERT INTO "studies" (
   "notes"
 ) VALUES (
   $(createdAt),
-  $(userSubject),
+  $(userId),
   $(studyRequestId),
   $(studyType),
   $(daysOfWeek),
@@ -71,17 +71,17 @@ INSERT INTO "studies" (
   $(notes)
 ) RETURNING "id"`;
     const createdAt = DateTime.local();
-    const { userSubject, id: studyRequestId } = studyRequest;
+    const { userId, id: studyRequestId } = studyRequest;
     const { id } = await db.one(sql, {
       createdAt,
-      userSubject,
+      userId,
       studyRequestId,
       ...study,
     });
     return {
       id,
       createdAt,
-      userSubject,
+      userId,
       studyRequestId,
       ...study,
     };
@@ -96,7 +96,7 @@ INSERT INTO "studies" (
   static async update(study) {
     const sql = `
 UPDATE "studies" SET
-  "userSubject" = $(userSubject),
+  "userId" = $(userId),
   "studyRequestId" = $(studyRequestId),
   "studyType" = $(studyType),
   "daysOfWeek" = $(daysOfWeek),
--- a/lib/db/UserDAO.js
+++ b/lib/db/UserDAO.js
@@ -18,6 +18,9 @@ class UserDAO {
    */
   static async byIds(ids) {
     const uniqueIds = Array.from(new Set(ids));
+    if (uniqueIds.length === 0) {
+      return new Map();
+    }
     const sql = 'SELECT * FROM "users" WHERE "id" IN ($(uniqueIds:csv))';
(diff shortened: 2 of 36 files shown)


lib/db/StudyDAO.js
```diff
@@ -53,7 +53,7 @@ SELECT
     const sql = `
 INSERT INTO "studies" (
   "createdAt",
-  "userSubject",
+  "userId",
```
candu: need to test these changes end-to-end.  can we add REST API tests for associated endpoints at some point?

lib/db/UserDAO.js
```diff
@@ -1,25 +1,31 @@
...
+import DateTime from '@/lib/time/DateTime';
 
 /**
  * Data access layer for users.
  */
 class UserDAO {
-  static async bySubject(subject) {
-    const sql = 'SELECT * FROM "users" WHERE "subject" = $(subject)';
-    return db.oneOrNone(sql, { subject });
+  static async byId(id) {
```
candu: add tests for `byId`, `byIds`, `byEmail`, etc. in `DAO.spec.js`.

lib/email/EmailStudyRequestConfirmation.js
```diff
@@ -45,11 +45,9 @@ class EmailStudyRequestConfirmation extends EmailBase {
   }
 
   getRecipients() {
-    const { email, name } = this.user;
+    const { email } = this.user;
```
candu: need to test email flows as well

web/views/FcAdfsCallback.vue
```diff
@@ -0,0 +1,83 @@
...
+      idToken,
+      nonce,
+    };
+  },
+  computed: {
+    ...mapState(['auth']),
+  },
+  mounted() {
+    if (this.idToken === null) {
+      // TODO: redirect back to login
```
candu: resolve these, please :)

web/views/FcLogin.vue
```diff
@@ -5,123 +5,56 @@
...
 </template>
 
 <script>
-import { email, required } from 'vuelidate/lib/validators';
 import { mapState } from 'vuex';
 
-import TdsPanel from '@/web/components/tds/TdsPanel.vue';
+const HEX = '0123456789abcdef';
+
+function randomNonce(length) {
```
candu: move this to a separate library, e.g. `@/lib/auth/Nonce`.

also, `length` is kind of confusing: is it the number of bits, hex digits, or bytes?  let's rename to `numBytes` for clarity.

lib/db/StudyDAO.js
candu: punting REST API tests here to future iterations

PR #296, opened 2020-01-10
Seamless login

### LINKED ISSUE #295: Seamless login
**Description**
We currently require all users to login immediately via a "gatekeeper" page that is shown up front.

In this task, we move towards a more user-friendly process where users can view the map and associated (non-private) data while logged out.  The user dropdown moves to the left sidebar; when logged out, it is replaced by a "Log In" button (where "Help" is now).

**Acceptance Criteria**
- [x] users can view the map, data, and reports sections of MOVE without logging in;
- [x] the top bar and "Help" button are removed;
- [x] the user dropdown has been moved to bottom left, replacing "Help";
- [x] when logged out, this dropdown is replaced by the "Log In" button;
- [x] logging in works as before;
- [ ] when visiting a path that requires authentication, the user is automatically redirected to login, and once login is completed they are redirected to their original destination.

candu:
This PR closes #295 by changing the way login works on MOVE from an up-front gatekeeper page to a "Log In" button at bottom-left.  The user can now access the View Map, View Data, and reporting parts of MOVE without logging in, and can log in from any of those pages by clicking "Log In".

It also starts down the path to Material Design: we now use [Vuetify](https://vuetifyjs.com/) for the "Log In" button, as well as for `SearchBarLocation`.  We've also replaced left navbar icons with [MDI](https://materialdesignicons.com/) equivalents.

At this point, Material Design integration is _very_ incomplete, and so these changes currently look a bit out of place.  Furthermore, we currently have the overhead of loading both our components / styles and Material Design components / styles.  There will be several follow-up issues to port over specific parts of the site.


candu:
Giving this a rubber stamp for now - we'll have several follow-up issues, and can get stricter about cleanup / review as this gets closer to release-worthy.

PR #306, opened 2020-01-30
Initial Material flows: View Map, View Data, Show Reports

### LINKED ISSUE #305: Implement designs: View Map (initial), View Data, Show Reports
**Description**
We now have higher-fidelity designs for some of the user flows in MOVE, courtesy of @shinechaudhuri .  This dovetails nicely with our gradual shift from alpha to beta, as part of which we've already built out a solid Material UI basis for MOVE over the last several PRs.

In this task, we implement some of the flows from our new designs on top of this basis, using standard Material components where possible and modifying said components where needed.

**Acceptance Criteria**
- [x] View Map light mode implemented;
- [x] View Data light mode implemented;
- [x] Show Reports light mode implemented.

**Additional Notes**
For now, we'll ignore light / dark mode switching - that can be integrated later.  These can be at a "first pass" fidelity for now, pending follow-up pairing around design details.

candu:
This PR closes #305 by implementing "first pass" versions of the View Map, View Data, and Show Reports mocks provided by @shinechaudhuri .


web/components/FcDataTableStudies.vue
```diff
@@ -0,0 +1,83 @@
...
+  computed: {
+    countSummaryOrLoading() {
+      if (this.loading) {
+        return [];
+      }
+      return this.countSummary;
+    },
+  },
+  methods: {
+    onActionShowReports(item) {
```
candu: just move the `$emit` up top into the template - no reason to have these one-liner actions

web/components/FcDisplayViewDataAtLocation.vue
```diff
@@ -326,10 +283,14 @@ export default {
     ...mapMutations('requestStudy', [
       'setNewStudyRequest',
     ]),
+    ...mapMutations('viewData', [
+      'removeFilter',
+      'setFilters',
+    ]),
     ...mapMutations([
+      'setDialog',
```
candu: can we get rid of `setDialog` altogether (here and elsewhere)?  we're moving away from that to more localized instantiation of dialogs, which helps reduce the amount of state we have to store in `vuex`.

web/components/FcDrawerViewReportsAtLocation.vue
```diff
@@ -0,0 +1,402 @@
...
+            icon
+            @click="actionNavigateBack">
+            <v-icon>mdi-chevron-left</v-icon>
+          </v-btn>
+          <h1 class="subtitle-1">Turning Movement Count</h1>
+          <div class="mx-1">&#x2022;</div>
+          <div>{{location.description}}</div>
+
+          <div
+            v-if="filterChips.length > 0">
```
candu: here we should take out all chips related to the `studyType` filter - maybe a computed property could help, similar to what we did with `filterParamsPaginated`?

web/components/PaneMap.vue
```diff
@@ -8,15 +8,15 @@
         indeterminate />
     </div>
     <SearchBarLocation
-      class="search-bar-location"
-      v-model="internalLocation" />
+      v-if="!drawerOpen"
+      :class="{
```
candu: something about this doesn't feel right: we're adding a `drawer-open` class (with CSS below), but the condition doesn't really have to do with the drawer being open, and in any case we hide this bar from `PaneMap` altogether when the drawer is open.

can we remove this class?  please test.

web/components/PaneMap.vue
```diff
@@ -728,39 +712,29 @@ export default {
...
     z-index: var(--z-index-controls);
   }
-  & > .search-bar-location {
-    top: var(--space-m);
+  & > .fc-search-bar-location {
+    top: 20px;
     position: absolute;
     left: var(--space-l);
     z-index: var(--z-index-controls);
+    &.drawer-open {
```
candu: same here: see if we can get rid of this

web/components/SearchBarLocation.vue
```diff
@@ -1,28 +1,29 @@
 <template>
   <v-autocomplete
     v-model="keystring"
+    append-icon="mdi-magnify"
```
candu: didn't realize that this causes the icon to flip around.  can we get it to stop doing that, or move this icon somewhere that it won't have this behaviour?

web/components/dialogs/FcDialogReportParameters.vue
```diff
@@ -0,0 +1,68 @@
...
+    value: Boolean,
+  },
+  data() {
+    const internalReportParameters = JSON.parse(JSON.stringify(this.reportParameters));
+    return {
+      internalReportParameters,
+    };
+  },
+  computed: {
+    internalValue: {
```
candu: i'd originally tried to extract this `value` / `internalValue` pattern into a Vue mixin, but hit a stumbling block: `props` need types.  (well, they don't _need_ them, but the extra type-checking is good!)

there *is* a solution, however!  we can create a "mixin factory": a module that exports a function `getMixin(type)`, which itself returns a mixin.  we can then call `getMixin` within the `mixins` array to create the mixin instance, since under the hood it's just doing fancy object merging.  (no idea what this pattern would look like under the upcoming Vue 3.x composition API, but we don't need to worry about that - they've promised backwards compatibility for this major version at least.)

we should probably do this in a separate "cleanup" PR.

web/components/dialogs/FcModalShowReports.vue
```diff
@@ -160,7 +160,6 @@ import { mapGetters } from 'vuex';
 import FcReport from '@/web/components/reports/FcReport.vue';
 import FcReportParametersWarrantTrafficSignalControl
   from '@/web/components/reports/FcReportParametersWarrantTrafficSignalControl.vue';
-import TdsMixinModal from '@/web/components/tds/TdsMixinModal';
```
candu: we should just get rid of `FcModalShowReports` altogether

web/components/tds/tds.postcss
```diff
@@ -91,6 +91,7 @@ body {
   margin: 0;
   min-height: 100vh;
   padding: 0;
+  overflow: hidden;
```
candu: not sure why this is necessary, but i was seeing an errant full-page scrollbar after the `FcLayoutDrawerMap` changes

web/router.js
```diff
@@ -145,31 +159,29 @@ async function beforeEachCheckAuth(to) {
...
-      return { name: 'home', query: { path, login: true } };
+      return loggedIn;
     }
     if (metaAuth === false) {
       // this route requires an unauthenticated user
-      return loggedIn ? { name: 'home' } : false;
+      return !loggedIn;
     }
-    return false;
+    return true;
```
candu: now that we have the `onError` handler in `router`, we shouldn't need to catch the error here - having more unified handling of this is a Good Thing

web/router.js
```diff
@@ -145,31 +159,29 @@ async function beforeEachCheckAuth(to) {
...
   }
 }
 
 router.beforeEach((to, from, next) => {
   beforeEachCheckAuth(to)
-    .then((redirect) => {
-      if (redirect) {
-        next(redirect);
-      } else {
+    .then((meetsAuthRequirements) => {
```
candu: rewrite in `async` format

web/router.js
```diff
@@ -191,4 +203,26 @@ router.afterEach((to) => {
   afterEachSetTitle(to);
 });
 
+function onErrorShowToast(err) {
+  // TODO: only show this for study request stuff!
```
candu: we should ensure that our backends are returning sensible errors here (but we can do that in a separate PR)

web/router.js
```diff
@@ -191,4 +203,26 @@ router.afterEach((to) => {
...
+    variant: 'error',
+    text: err.message,
+  };
+}
+
+router.onError((err) => {
+  const { currentRoute } = router;
+  const toast = onErrorShowToast(err, currentRoute);
+  if (toast) {
+    store.dispatch('setToast', false);
```
candu: this should be `store.dispatch('setToast', toast)`

web/store/index.js
```diff
@@ -26,8 +31,8 @@ export default new Vuex.Store({
     now: DateTime.local(),
     requestReasons: [],
     // TOP-LEVEL UI
+    dialog: null,
```
candu: get rid of this and other `dialog` / `modal`-related stuff in here, please

web/views/FcLayoutDrawerMap.vue
```diff
@@ -1,20 +1,55 @@
...
+    }">
+    <div
+      v-if="hasDrawer"
+      class="pane-drawer-toggle elevation-3 font-size-xl"
+      @click="setDrawerOpen(!drawerOpen)">
+      <div class="text-center">
+        <v-icon
+          v-if="vertical"
+          small>
+          {{drawerOpen ? 'mdi-menu-down' : 'mdi-menu-up'}}
```
candu: can we just extract this into a computed property and have something like `<v-icon small>{{drawerToggleIcon}}</v-icon>`?

web/views/FcLayoutDrawerMap.vue
```diff
@@ -1,20 +1,55 @@
...
+        </v-icon>
+      </div>
+    </div>
+    <div
+      class="fc-pane-wrapper d-flex fill-height"
+      :class="{
+        'flex-column': vertical,
+      }">
+      <div
+        v-show="(hasDrawer && drawerOpen) || vertical"
```
candu: same here: extract into `showDrawer`

web/views/FcLayoutDrawerMap.vue
```diff
@@ -1,20 +1,55 @@
...
+        class="elevation-3 flex-grow-1 flex-shrink-0"
+        :class="{
+          'order-2': vertical,
+        }">
+        <router-view></router-view>
+      </div>
+      <div
+        v-show="!drawerOpen || !vertical"
```
candu: ...and `showMap`

web/views/FcLayoutDrawerMap.vue
```diff
@@ -24,7 +59,78 @@ export default {
...
+.fc-layout-drawer-map {
+  position: relative;
+  width: 100%;
+
+  & > .fc-pane-wrapper > div {
+    flex-basis: 0;
+  }
+
+  & > .pane-drawer-toggle {
+    background-color: var(--white);
```
candu: make this appear visually "behind" the drawer, if possible

lib/db/PoiDAO.js
```diff
@@ -0,0 +1,63 @@
...
+import db from '@/lib/db/db';
+import { InvalidCentrelineTypeError } from '@/lib/error/MoveErrors';
+
+const SCHOOL_RADIUS = 250;
+
+class PoiDAO {
+  // SEGMENT
+  static async bySegmentSchool(centrelineId) {
+    const sql = `SELECT s.objectid AS id, ST_Distance(
+      ST_Transform(s.geom, 26917),
```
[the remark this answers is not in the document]
candu: going to recommend that we do this in a follow-up PR

web/components/SearchBarLocation.vue
candu: found the appropriate selector and added it here :)

Issue #309, opened 2020-02-04
Constants cleanup


candu:
**Description**
We have several enum-like values in `@/lib/Constants` left over from early prototyping.  Now that we have `Enum`, however, it would make more sense to use that consistently.

**Acceptance Criteria**
The following are addressed:

- [x] `CountHours` (replace with `StudyHours`)
- [x] `COUNT_TYPES` (rename to `StudyType`)
- [x] `HttpStatus` (can use `enumValueOf(statusCode, 'statusCode')` and make `isOk` a method?)

**Additional Notes**
In many cases, these will involve changes in both frontend and backend, especially around the validation of schemas.

PR #311, opened 2020-02-04
Material rework: Request Management

### LINKED ISSUE #307: Material rework: Request Study, Track Requests, View Request
**Description**
Building on #305 , this task moves on to implementation of Request Study and Track Requests as designed by @shinechaudhuri.

**Acceptance Criteria**
- [x] Request Study: form fields;
- [x] Request Study: hint copy;
- [x] Request Study: validation error copy;
- [x] Request Study: request submission;
- [x] Track Requests: tab bar at top;
- [x] Track Requests: table layout;
- [x] Track Requests: individual actions;
- [x] Track Requests: multiple-selection actions;
- [x] View Request: summary view of requests;
- [x] View Request: comments;
- [x] View Data: show studies in progress below table.

**Additional Notes**
This is likely to be another sizable "sweeping change" PR, but should *hopefully* be the last.

candu:
This PR closes #307 by converting the Request Study, Track Requests, and View Request flows to Material Design.

In the process, we make some small changes to the `study_requests` schema to provide functionality required in the new designs.


lib/api/WebApi.js
```diff
@@ -137,19 +155,34 @@ async function getUsersByIds(ids) {
     return new Map();
   }
   const options = { data: { id: idsArray } };
-  const users = await apiClient.fetch('/users/byId', options);
+  let users = await apiClient.fetch('/users/byId', options);
+
+  const usersSchema = Joi.array().items(
```
candu: ideally we'd just share these schema definitions directly between frontend and backend - but that can wait.

lib/config/MoveConfig.js
```diff
@@ -103,8 +103,8 @@ const configSchema = Joi.object().keys({
   }),
   host: Joi.string().hostname(),
   https: Joi.object().keys({
-    key: Joi.binary().required(),
-    cert: Joi.binary().required(),
+    key: Joi.any().required(),
```
candu: ...because we're now using the browser build in both client and server, which doesn't support `.binary()` (because that depends on the node.js `Buffer` type, IIRC).

please document this here.

lib/controller/WebInitController.js
```diff
@@ -1,32 +0,0 @@
-import StudyRequestReasonDAO from '@/lib/db/StudyRequestReasonDAO';
```
candu: ...we might actually still want this around `study_request_assignees`, since those *should* be dynamic.

web/components/FcDrawerViewData.vue
```diff
@@ -102,6 +102,19 @@
...
+          <div class="pa-5">
+            <div
+              v-for="study in studiesPending"
+              :key="study.id"
+              class="align-center d-flex">
+              <v-icon
+                color="warning"
+                left>mdi-information</v-icon>
+              <div>
+                {{study.studyType}} has been requested on {{study.createdAt | date}}.
```
candu: show the human-readable name.  (this can wait until #309 , since that should make this change easier.)

web/components/inputs/FcCheckboxGroupChips.vue
```diff
@@ -0,0 +1,53 @@
...
+    },
+    itemValue: {
+      type: String,
+      default: 'value',
+    },
+    items: Array,
+    value: Array,
+  },
+  computed: {
+    internalValue: {
```
candu: use `FcMixinVModelProxy` here

web/components/inputs/FcRadioGroup.vue
```diff
@@ -0,0 +1,48 @@
...
+    </template>
+  </v-radio-group>
+</template>
+
+<script>
+import FcMixinVModelProxy from '@/web/mixins/FcMixinVModelProxy';
+
+export default {
+  name: 'FcRadioGroup',
+  mixins: [FcMixinVModelProxy([Boolean, Number, Object, String])],
```
candu: let's add an untyped version of `FcMixinVModelProxy` - if we don't pass in the `type` parameter, it should just use a [prop string array](https://vuejs.org/v2/guide/components-props.html) to add the `value` property.

web/mixins/FcMixinRouteAsync.js
```diff
@@ -4,10 +4,8 @@ export default {
       loading: true,
     };
   },
-  beforeRouteEnter(to, from, next) {
-    next((vm) => {
-      vm.loadAsync(to);
-    });
+  created() {
+    this.loadAsync(this.$route);
```
candu: document this: why `created` instead of `beforeRouteEnter`?

web/mixins/FcMixinVModelProxy.js
```diff
@@ -0,0 +1,15 @@
+export default type => ({
```
candu: as mentioned above: please add typeless version

web/store/index.js
```diff
@@ -144,7 +141,9 @@ export default new Vuex.Store({
         csrf: state.auth.csrf,
         data,
       };
-      return apiFetch(url, options);
+      const studyRequestSaved = apiFetch(url, options);
```
candu: while this will still work (the `async` function is still returning a promise), it's a bit misleading: we might expect `studyRequestSaved` to be the study request, not a `Promise` that resolves to the study request.

since it's still a good idea to give quick user-facing feedback, I'd suggest:

```js
commit('setToast', REQUEST_STUDY_SUBMITTED);
return apiFetch(url, options);
```

lib/controller/WebInitController.js
candu: (going to fix in a later revision - need to ask @shinechaudhuri how assignment is being handled here)

PR #338, opened 2020-03-02
Data pipeline MATERIALIZED VIEW revamp

### LINKED ISSUE #334: Update MOVE data pipelines to use MATERIALIZED VIEW / TRUNCATE
**Description**
Our current data pipelines perform a full `DROP TABLE` / `CREATE TABLE AS (SELECT ...)` cycle every time they execute.  We should migrate to `REFRESH MATERIALIZED VIEW CONCURRENTLY` or `TRUNCATE` / `INSERT INTO SELECT` where possible.

**Acceptance Criteria**
These jobs have been migrated:
- [x] CRASH normalization and geocoding;
- [x] FLOW normalization and geocoding;
- [x] artery intersections;
- [x] ArcGIS layer fetching.

candu:
This PR closes #334 by updating our MOVE data pipelines in Airflow to use `TRUNCATE` / `REFRESH MATERIALIZED VIEW CONCURRENTLY` instead of dropping tables and re-creating them each time.

Using `MATERIALIZED VIEW`s in this way has several advantages:

- updates are now zero-downtime;
- `REFRESH MATERIALIZED VIEW CONCURRENTLY` gives us incremental updates for free (by diffing between old and new copies internally);
- downstream jobs can now use `VIEW`s.

One tradeoff is the possibility of [table bloat](https://stackoverflow.com/questions/52421411/refreshing-materialized-view-concurrently-causes-table-bloat) over time.  However, Amazon RDS instances are configured to [autovacuum by default](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.html#Appendix.PostgreSQL.CommonDBATasks.Autovacuum), so this should not be an issue.

Another tradeoff is that schema changes are now a bit harder to make, and will likely require more downtime than before.


scripts/airflow/dags/copy_opendata_shapefiles.py
```diff
@@ -0,0 +1,44 @@
...
+
+from airflow.operators.bash_operator import BashOperator
+
+from airflow_utils import create_dag
+
+START_DATE = datetime(2020, 2, 27)
+SCHEDULE_INTERVAL = '30 4 * * 6'
+DAG = create_dag(__file__, __doc__, START_DATE, SCHEDULE_INTERVAL)
+
+TASKS = {
```
candu: can we add a comment here explaining where these seemingly random UUIDs come from?

scripts/airflow/tasks/build_artery_centreline/build_artery_centreline.sql
```diff
@@ -1,26 +1,42 @@
...
+        2 as centreline_type,
+        ST_SetSrid(ST_MakePoint(longitude, latitude), 4326) as geom
+      from prj_volume.artery_intersections
+    )
   )
+  SELECT
+    arterycode,
+    centreline_id,
+    centreline_type,
+    MODE() WITHIN GROUP(ORDER BY geom) AS geom
```
candu: we should really be using `MODE()` for collisions normalization as well.  (doesn't need to be fixed right away, just flagging it here.)

scripts/airflow/tasks/build_artery_centreline/build_artery_centreline.sql
```diff
@@ -1,26 +1,42 @@
...
 );
-create index artery_centreline_arterycode on prj_volume.artery_centreline (arterycode);
-create index artery_centreline_centreline_id on prj_volume.artery_centreline (centreline_id);
-create index artery_centreline_geom on prj_volume.artery_centreline using gist (geom);
-create index artery_centreline_srid3857_geom ON prj_volume.artery_centreline using gist (ST_Transform(geom, 3857));
+create unique index IF NOT EXISTS artery_centreline_arterycode on prj_volume.artery_centreline (arterycode);
+create index IF NOT EXISTS artery_centreline_centreline on prj_volume.artery_centreline (centreline_type, centreline_id);
+create index IF NOT EXISTS artery_centreline_geom on prj_volume.artery_centreline using gist (geom);
+create index IF NOT EXISTS artery_centreline_srid3857_geom ON prj_volume.artery_centreline using gist (ST_Transform(geom, 3857));
+create index IF NOT EXISTS artery_centreline_srid2952_geom ON prj_volume.artery_centreline using gist (ST_Transform(geom, 2952));
```
candu: note that `PoiDAO` still uses the old SRID 26917 projection - please change in follow-up work.

Issue #394, opened 2020-04-07
Implement role-based access control


candu:
**Description**
MOVE currently understands three "roles": anonymous, staff, and supervisor.  The supervisor role is managed not via user metadata, but via a GET query parameter on certain routes.  This setup is best understood as more of an ad-hoc prototype than a robust RBAC approach.

This task seeks to implement actual RBAC by defining user roles and permissions, by checking those roles and permissions across the application, and by offering UI and/or script tools for RBAC administration.

**Acceptance Criteria**
Roughly broken down into functional PRs:

- [x] First Steps
  - [x] define role-permission matrix;
  - [x] investigate RBAC implementation / administration approaches;
- [x] Database Layer
  - [x] define `AuthScope` enum in `@/lib/Constants`;
  - [x] add `scope` column to users table via new migration, and give `studyRequests` to all existing users;
  - [x] add scopes to user objects in `UserDAO` and `AuthController`, and give `studyRequests` to all new users;
- [x] Scope Configuration
  - [x] configure required scopes for REST API endpoints in `StudyController`, `StudyRequestController`;
  - [x] implement `scope` getter in Vuex store;
  - [x] configure required scopes for `vue-router` routes in frontend, and integrate with `beforeEach` hook;
- [ ] Admin Panel
  - [x] implement RBAC admin DAO layer and endpoints;
  - [ ] implement RBAC admin UI, and give someone `admin` access;
- [x] Dynamic Checks
  - [x] implement `hasAuthScope` function, e.g. in new module within `@/lib/auth`;
  - [x] implement dynamic `hasAuthScope` checks:
    - [x] `StudyRequestController.putStudyRequest` to gate assignment and status updates;
    - [x] frontend navbar, to hide RBAC admin panel for non-admin users;
    - [x] others?

This list deliberately leaves necessary DAO and REST API testing out, as those are captured by #381 and #385.  Nevertheless, it might be a good idea to start adding tests as we go along.

**Additional Notes**
There's the [`hapi-rbac`](https://www.npmjs.com/package/hapi-rbac) package, but that appears to be unmaintained at this point.

Hapi itself has [scopes](https://futurestud.io/tutorials/hapi-restrict-user-access-with-scopes), which appear to borrow from the OAuth concept of the same name.  Routes can be configured to require / forbid certain scopes, which means these can act as permissions.

PR #396, opened 2020-04-08
RBAC user scopes: migration, creation, DAO updates

candu:
# Issue Addressed
This PR makes further progress on #394 - still several steps to go!

# Description
This implements the database layer for user authentication scopes, and beefs up testing / validation a tiny bit around this at the same time.

# Tests
`npm run test:test-db` passes.  I haven't checked that `GET /api/auth` works still; that will be tested in subsequent PRs alongside other REST API / frontend-facing changes.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -2,6 +2,18 @@ import { Enum } from '@/lib/ClassUtils';
 import TimeFormatters from '@/lib/time/TimeFormatters';
 
 /**
+ * Authentication scopes.  Used to restrict access to certain REST API endpoints,
+ * frontend routes, and features.
+ */
+class AuthScope extends Enum {}
+AuthScope.init([
+  'ADMIN',
+  'STUDY_REQUESTS',
+  'STUDY_REQUESTS_ADMIN',
+  'STUDY_REQUESTS_EDIT',
+]);
+
+/**
  * Four cardinal directions.  Our legacy FLOW system stores Turning Movement Count
  * data according to these directions.  Additionally, since Toronto's streets operate
  * on a grid system, it is often useful to know cardinal directions of travel.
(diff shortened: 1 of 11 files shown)


lib/Constants.js
```diff
@@ -1,6 +1,18 @@
 import { Enum } from '@/lib/ClassUtils';
 import TimeFormatters from '@/lib/time/TimeFormatters';
 
+/**
+ * Authentication scopes.  Used to restrict access to certain REST API endpoints,
+ * frontend routes, and features.
+ */
+class AuthScope extends Enum {}
+AuthScope.init([
```
candu, line 9: this actually seems better than having a separate `scope` property 👍

lib/controller/AuthController.js
```diff
@@ -104,6 +108,9 @@ AuthController.push({
   path: '/auth',
   options: {
     auth: { mode: 'try' },
+    response: {
```
candu, line 111: can we at least test this in the frontend quickly?  might as well, since it's easier to correct at this stage

lib/db/UserDAO.js
```diff
@@ -22,44 +37,57 @@ class UserDAO {
...
+  "sub",
+  "uniqueName"
+) VALUES (
+  $(createdAt),
+  $(email),
+  $(scope),
+  $(sub),
+  $(uniqueName)
+) RETURNING "id"`;
+    const persistedUser = {
```
candu, line 72: I like that we're landing on a reliable CRUD pattern that incorporates both the transient vs. persisted language and Joi validation - makes these DAOs a lot more readable.

lib/db/UserDAO.js
```diff
@@ -22,44 +37,57 @@ class UserDAO {
...
+    return normalizeUser(persistedUser);
   }
 
   static async update(user) {
     const sql = `
 UPDATE "users"
-  SET "email" = $(email), "uniqueName" = $(uniqueName)
+  SET
+    "email" = $(email),
+    "scope" = $(scope),
```
candu, line 86: can we add a test for this in `DAO.spec.js`?

Issue #385, opened 2020-03-27
Testing coverage: REST APIs


candu:
**Description**
We currently have extremely minimal testing on our REST APIs, and absolutely no coverage targets.  This will have to change, especially as we start to work with role-based access controls.

**Acceptance Criteria**
- [x] remove `lib/controller` from coverage exclusion in `jest.config.js`;
- [ ] test controllers where possible:
  - [ ] `AuthController`
  - [ ] `CollisionController`
  - [ ] `CountController`
  - [ ] `DynamicTileController`
  - [ ] `LocationController`
  - [ ] `PoiController`
  - [ ] `SignalSuggestionController`
  - [ ] `StudyController`
  - [ ] `StudyRequestController`
  - [ ] `UserController`
- [ ] add tests for request / response type validation using `Joi`
- [ ] add tests for authenticated vs. anonymous access
- [ ] add tests for role-based access (where needed)

**Additional Notes**
There's [`server.inject()`](https://livebook.manning.com/book/hapi-js-in-action/chapter-10/) for testing Hapi servers.  We can also build on our current framework, which actually starts up a live server in test mode and issues actual HTTP requests to it.  IMHO using a live server is preferable, as this is closest to how the server is used in the actual web application.

In either case, we may need to extend `/api/auth/test-login` in `AuthController` to specify access roles for the given user.

One issue here is that we'll be duplicating a lot of work from `DAO.spec.js` - in many cases, responses are just thin wrappers around DAO method return values.

To extend coverage into `WebApi`, we can call endpoints through there where possible and add a `fetch` polyfill` to `tests/unitSetup`.

PR #397, opened 2020-04-08
RBAC: scope configuration, dynamic matching, stub admin UI

candu:
# Issue Addressed
This PR makes significant further progress on #394 .

# Description
Several pieces coming together:

- added scopes to all backend REST API endpoints and frontend routes;
- added REST API endpoints and DAO methods to get all users, and to update a user object;
- added `@/lib/auth/ScopeMatcher` to perform dynamic checks;
- added '/admin` routes with stub UI implementation, including a second tab that will eventually be used for site-wide metrics;
- added `FcDialogConfirmUnauthorized`, which is used to notify the user when they attempt to access a feature to which they do not currently have access.

More on this last one: while outright hiding features can be better in some cases, there's always the possibility that we'll miss something, so it helps to have a user-facing way to remediate.

# Tests
- tested "Request Study" using the `esavage` user;
- modified permissions in the database to exercise authorized vs. unauthorized paths;
- tested stub admin UI manually;
- `npm run test:test-db` to test DAO changes.

This still requires more extensive REST API testing, as specified in #385 .

### DIFF
--- a/lib/controller/StudyController.js
+++ b/lib/controller/StudyController.js
@@ -1,8 +1,8 @@
 import Joi from '@/lib/model/Joi';
 
+import { AuthScope, CentrelineType } from '@/lib/Constants';
 import StudyDAO from '@/lib/db/StudyDAO';
 import Study from '@/lib/model/Study';
-import { CentrelineType } from '@/lib/Constants';
 
 /**
  * Routes related to metadata for studies submitted via MOVE.  Note that this is different from
@@ -26,6 +26,9 @@ StudyController.push({
   method: 'GET',
   path: '/studies/byCentreline/pending',
   options: {
+    auth: {
+      scope: [AuthScope.STUDY_REQUESTS.name],
+    },
     response: {
       schema: Joi.array().items(Study.read),
     },
--- a/lib/controller/StudyRequestController.js
+++ b/lib/controller/StudyRequestController.js
@@ -1,7 +1,7 @@
 import Boom from '@hapi/boom';
 import Joi from '@/lib/model/Joi';
 
-import { StudyRequestStatus } from '@/lib/Constants';
+import { AuthScope, StudyRequestStatus } from '@/lib/Constants';
 import StudyRequestDAO from '@/lib/db/StudyRequestDAO';
 import StudyRequestCommentDAO from '@/lib/db/StudyRequestCommentDAO';
 import EmailStudyRequestConfirmation from '@/lib/email/EmailStudyRequestConfirmation';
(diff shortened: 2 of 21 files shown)


lib/controller/StudyRequestController.js
```diff
@@ -139,23 +137,23 @@ StudyRequestController.push({
...
     },
     validate: {
       params: {
         id: Joi.number().integer().positive().required(),
       },
-      payload: StudyRequest.update.append({
-        // TODO: remove when we have RBAC
-        isSupervisor: Joi.boolean().default(false),
-      }),
+      payload: StudyRequest.update,
```
candu, line 150: this is the only use of `StudyRequest.update` - but in this case, we need the distinction with `StudyRequest.read`, as a study request being updated might contain both new studies and previously existing studies.

longer-term, we should probably standardize this pattern across our models.  one possible approach: define a `Model` class that includes one or more `ModelColumn` objects.  a `ModelColumn` is:

- a flag that determines whether this is a transient or persisted column;
- one of:
  - a Joi schema used to validate this column;
  - another `Model` (X-to-one relationships);
  - an array containing a `Model` as its only element (X-to-many relationships).

note that this does not manage any SQL / CRUD stuff - this is *only* about validation, though we could add a lightweight CRUD layer.

anyways - definitely out of scope here, but some food for thought.

web/router.js
```diff
@@ -10,16 +12,54 @@ const router = new Router({
...
+          scope: [AuthScope.ADMIN],
+        },
+        title: 'Admin Console',
+      },
+      component: () => import(/* webpackChunkName: "admin" */ '@/web/views/FcAdmin.vue'),
+      children: [{
+        path: '',
+        name: 'admin',
+        meta: {
+          auth: {
```
candu, line 28: I *think* we can get rid of the `meta` declarations on children here and just delegate to the parent route - let's try that out.  (if we can't, I'd say we should fix that instead of duplicating scopes on every child route!)

lib/controller/StudyController.js
```diff
@@ -26,6 +26,9 @@ StudyController.push({
   method: 'GET',
   path: '/studies/byCentreline/pending',
   options: {
+    auth: {
+      scope: [AuthScope.STUDY_REQUESTS.name],
```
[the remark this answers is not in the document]
candu: answer: no; Hapi's route schemas are validated by Joi, which doesn't accept `AuthScope` instances instead of strings here.

however, it *does* correctly match the `AuthScope` objects stored in `user.scope` against these, so that's good at least.

web/router.js
candu: yup, we can get rid of these 🎉

PR #412, opened 2020-04-16
One study per request

### LINKED ISSUE #403: Request Management: One study per request
**Description**
We currently allow users to include multiple study types in a single request.  However, this complicates both Study Request and Track Requests without delivering much additional value to users.

This task addresses this by changing study requests to only include exactly one study.

**Acceptance Criteria**
- [x] change schema to only link one study to a request;
- [x] update `StudyRequest` and `Study` models accordingly;
- [x] update DAOs and associated tests;
- [x] update relevant REST API endpoints;
- [x] update `WebApi` methods and associated callsites;
- [x] update Request Study to use radio buttons instead of chips for selecting study type;
- [x] update View Request to show info for single study;
- [x] add Type column to Track Requests.
- [x] ensure that relevant tests pass, and that relevant user-facing flows still work.

**Additional Notes**
There are a few possible strategies for changing the `study_requests` / `studies` relationship:

1. inline all `studies` fields in `study_requests`, and get rid of `studies` as a separate thing;
2. add a `studyId` field in `study_requests` to enforce a 1-to-1 relationship;
3. keep schema as-is, and just validate / normalize as needed in DAOs.

1) reduces schema / codebase complexity, while 3) is probably the least additional work.  1) and 2) best maintain DB schema normalization.

This also has implications for reconciling incoming counts to study requests later.

IMHO we should go with 1).  We can later create a view `counts.studies` that combines `TRAFFIC.COUNTINFO`, `TRAFFIC.COUNTINFOMICS` and `study_requests` into a single view linking data to locations and study requests.  This would still be possible under 2) or 3), but we'd essentially be duplicating a lot of info between `studies` and `counts.studies`.

1) also simplifies the process of bulk study request creation: just create a "study request template" (e.g. request this kind of study at all locations in this selected area that don't have them), then apply to each location.  2) or 3) would require multi-stage setup for this, which is more brittle.

candu:
# Issue Addressed
This PR closes #403 .

# Description
Changes the database schema, web application backend, and frontend user flows to switch over to one study per request.

# Tests
- updated DAO tests and ran `npm run test:test-db` successfully;
- tested several user flows manually: Request Study, email notification, View Request, Track Requests, export requests to CSV.


scripts/db/schema-11.up.sql
```diff
@@ -0,0 +1,15 @@
+BEGIN;
+
+DROP TABLE "studies";
```
candu, line 3: we can get away with this right now, but as we move into QA / staging / production we'll need to be much more diligent about preserving user data across migrations.

(just something to think about, nothing actionable here right now.)

web/views/FcRequestsTrack.vue
```diff
@@ -397,7 +380,7 @@ export default {
       });
     },
     actionDownload(studyRequests) {
-      const rows = Array.prototype.concat.apply([], studyRequests.map(getItemRows));
+      const rows = Array.prototype.concat.apply([], studyRequests.map(getItemRow));
```
candu, line 383: just do `const rows = studyRequests.map(getItemRow);` - we're no longer returning multiple rows per study request (another thing that becomes simpler / clearer!)

PR #414, opened 2020-04-20
Track requests UI revamp

### LINKED ISSUE #405: Request Management: Track Requests UI revamp
**Description**
As in #404 , we also have [new Track Request designs](https://www.figma.com/file/w9fpc0UCJQ95efLGQ79aWR/v1.1---Track-Requests?node-id=0%3A1).

This task implements the designs, while (again) leaving larger functional pieces to their own follow-up tasks.

**Acceptance Criteria**
- [x] general layout: `h1`, off-white background, table card;
- [x] add chips stub at top-left;
- [x] add search bar stub at top-right;
- [x] fixed / scroll strategy for table card;
- [x] add Date Created, Assigned, Last Updated columns.

candu:
# Issue Addressed
This PR closes #405 .

# Description
Revamps the UI of Track Requests to include a header with chips and search bar, plus a card wrapping the requests table that fixes the action bar and table headers.

# Tests
- updated `Constants.spec.js` to match changes and ran it via `npx jest`;
- manual testing in browser.

### DIFF
--- a/web/components/inputs/FcSearchBarRequests.vue
+++ b/web/components/inputs/FcSearchBarRequests.vue
@@ -55,9 +55,8 @@ export default {
   }
 
   & > .v-input__prepend-outer {
-    margin-right: 0;
+    margin: 0 !important;
     & > .v-input.fc-search-bar-requests-column {
-      margin-top: -8px !important;
       & > .v-input__control {
         border-bottom-right-radius: 0;
         border-top-right-radius: 0;
--- a/web/components/inputs/FcSearchBarRequests.vue
+++ b/web/components/inputs/FcSearchBarRequests.vue
@@ -62,6 +62,11 @@ export default {
         border-top-right-radius: 0;
         width: 140px;
       }
+
+      &:not(.v-input--is-focused) fieldset {
+        border-right: 0;
+        border-width: 1px;
+      }
     }
   }
 }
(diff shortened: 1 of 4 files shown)


web/components/inputs/FcSearchBarRequests.vue
```diff
@@ -0,0 +1,68 @@
...
+        { text: 'All Columns', value: null },
+        ...searchableColumns,
+      ];
+    },
+  },
+};
+</script>
+
+<style lang="scss">
+.fc-search-bar-requests {
```
candu, line 48: can we take one more stab at dealing with this?  it looks like the `fieldset` manages the border here, which kinda makes sense if you think about `<v-input>` as a container for one or more closely related input controls.

web/components/inputs/FcSearchBarRequests.vue
```diff
@@ -0,0 +1,68 @@
...
+
+<script>
+import FcMixinVModelProxy from '@/web/mixins/FcMixinVModelProxy';
+
+export default {
+  name: 'FcSearchBarRequests',
+  mixins: [FcMixinVModelProxy(Object)],
+  props: {
+    columns: Array,
+    searchKeys: Object,
```
candu, line 31: this setup makes this viable as a generic search bar control for `FcDataTable`, as we're just reusing the same `columns` and `searchKeys`.

(that said: we should definitely not refactor it until we come across a second use case.)

PR #427, opened 2020-04-28
Initial pass at REST API tests

### LINKED ISSUE #385: Testing coverage: REST APIs

candu:
# Issue Addressed
This PR closes #385 .

# Description
We use hapi's `server.inject()` to directly issue requests to the MOVE backend server, without going through the HTTP stack.

In the process, we also remove the exclusion of `lib/controllers` from `jest` coverage metrics, and add enough REST API tests to exceed the current coverage target of 60%.  (This effort includes a few complementary DAO tests.)

Note that the coverage of `lib/controller` here is only about 40%.  This is OK for now, especially considering our previous coverage was close to 0% - plus we get a solid base on which to build out further REST API tests in the future.  The setup with `InjectBackendClient` also suggests a way to test `WebApi` - we just need to refactor that a bit to be able to mock out the `BackendClient` instance.

# Tests
- `npm run test:test-api`
- `npm run ci:jest-coverage`
- verify that MOVE frontend / backend launch without errors after the refactor
- quick in-browser manual "smoke test" to double-check that changes do not break MOVE.


lib/controller/AuthController.js
```diff
@@ -148,30 +145,4 @@ AuthController.push({
   },
 });
 
-/**
```
candu, line 151: 🎉

tests/jest/api/RestApi.spec.js
```diff
@@ -1,19 +1,675 @@
...
+import UserDAO from '@/lib/db/UserDAO';
+import AuthState from '@/lib/model/AuthState';
+import DAOTestUtils from '@/lib/test/DAOTestUtils';
+import { generateUser } from '@/lib/test/random/UserGenerator';
+import DateTime from '@/lib/time/DateTime';
+import { initialize } from '@/web/MoveServer';
 
-beforeAll(RestApiTestUtils.startup, RestApiTestUtils.TIMEOUT);
-afterAll(RestApiTestUtils.shutdown, RestApiTestUtils.TIMEOUT);
+class InjectBackendClient {
```
candu, line 18: we can leave this here for now, but we should eventually move this to its own file in `lib/test` and add tests.

tests/jest/api/RestApi.spec.js
```diff
@@ -1,19 +1,675 @@
...
+  // supervisor can update study request
+  client.setUser(supervisor);
+  response = await client.fetch(`/requests/study/${persistedStudyRequest.id}`, {
+    method: 'PUT',
+    data: persistedStudyRequest,
+  });
+  expect(response.statusCode).toBe(HttpStatus.OK.statusCode);
+  expect(response.result.studyRequestChange).toBeNull();
+  persistedStudyRequest = response.result.studyRequest;
+
```
candu, line 670: we're actually missing a _lot_ of the branches in `PUT /requests/study/:id`, but those can be added later.

Issue #358, opened 2020-03-19
Port arterycode conflation over to Airflow


candu:
**Description**
Our data pipelines still depend on `prj_volume.artery_tcl`, but we're currently using an old copy of this.

Unfortunately, since Flow Load uses arterycodes to set location, we have to deal with arterycodes at least until we can revamp that intake pipeline.  That's at least a few months out at this point.

In the interim, we'll need some way to conflate arterycodes with whatever our spatial referencing layer is (currently Toronto Centreline).  By understanding the current `prj_volume.artery_tcl` process, we can port parts of it over to Airflow, and see if there are ways to generalize it to be as referencing system-agnostic as possible.  This also allows us to keep this piece up to date while that discussion is ongoing.

**Acceptance Criteria**
- [x] get script (or link to it) from @aharpalaniTO ;
- [ ] port over to Airflow job:
  - [x] assign locations to nodes, and attempt conflation against `centreline_intersection`:
    - [x] match `LINK_ID` to `int_id`;
    - [x] if not, spatial match on `(X_COORD, Y_COORD)` to `geom`;
    - [x] if not, use `(X_COORD, Y_COORD)` and mark `int_id = NULL`;
  - [ ] assign locations to arteries, and attempt using nodes to conflate against centreline:
    - [ ] intersections:
      - [ ] (S07) match `LINKID` to `NODE.LINK_ID`, mark `int_id = NULL` otherwise;
      - [ ] (S07) infer directionality;
    - [ ] midblocks:
      - [ ] match from / to `LINKID` against `NODE.LINK_ID`, mark `int_id = NULL` otherwise;
      - [ ] if both from / to non-null, try to match `int_id` values against endpoints of `geo_id` midblock:
        - [ ] (S02) if one such `geo_id`, use that;
        - [ ] (S02) if multiple such `geo_id`, use text-search to find a best match among them;
        - [ ] (S05) if no such `geo_id`, use geometric approaches to find a best match among all midblocks;
      - [ ] (S06) if one of from / to non-null, use geometric approaches to find a best match among all midblocks;
      - [ ] otherwise, fail to match;
    - [ ] (S08, S09) apply manual corrections;
    - [ ] (S10) address short segment errors;
    - [ ] (S11) flip geometries / directions where needed.

**Additional Notes**
@aharpalaniTO mentioned that there are some manual steps in this process.  Depending on what proportion of arterycodes require manual linkage, these can be ported over in a number of ways:

[...]

PR #436, opened 2020-05-05
Arterycode conflation: step-by-step script framework

candu:
# Issue Addressed
This PR makes significant progress on #358 .

# Description
We add step-by-step scripts for arterycode conflation.  These are based on [previous `artery_tcl` scripts](https://github.com/CityofToronto/bdit_volumes/tree/master/volume_project/flow_data_processing/arterycode_mapping), with a few key changes:

- these are intended for use in an (upcoming) Airflow DAG, so that we can automatically schedule conflation runs;
- we separate intersection and midblock pipelines for readability;
- we place all these scripts in a framework that allows for later improvements.

# Tests
Ran each script manually in sequence - planning to test end-to-end via Airflow in a separate PR.


scripts/airflow/tasks/conflate_artery_centreline/A1_centreline_filtered.sql
```diff
@@ -0,0 +1,31 @@
...
+CREATE MATERIALIZED VIEW IF NOT EXISTS counts.centreline AS (
+  SELECT
+    geo_id::bigint AS geo_id,
+    fnode::bigint AS from_int_id,
+    tnode::bigint AS to_int_id,
+    geom
+  FROM gis.centreline
+  WHERE fcode < 202000
+);
+CREATE UNIQUE INDEX IF NOT EXISTS pvt_centreline_geo_id ON counts.centreline (geo_id);
```
candu, line 12: get rid of `pvt_` stuff

scripts/airflow/tasks/conflate_artery_centreline/C1_arteries_midblock_manual_corr.sql
```diff
@@ -0,0 +1,14 @@
+CREATE SCHEMA IF NOT EXISTS counts;
+
+-- TODO: figure out how / whether to handle midblock corrections here
```
candu, line 3: get rid of this line

scripts/airflow/tasks/conflate_artery_centreline/B1_arteries_intersection_manual_corr.sql
```diff
@@ -0,0 +1,11 @@
+CREATE SCHEMA IF NOT EXISTS counts;
+
+-- TODO: figure out how / whether to handle intersection corrections here
```
candu, line 3: can we at least take a stab at this?

IIRC main difficulty is that we need to reverse-engineer the intersection ID from the neighbouring segments, which is what @aharpalaniTO 's process in our existing `flow_geocoding` DAG does.  if we can just drop that in, we should do that!

PR #453, opened 2020-05-14
Collisions end-to-end filtering

### LINKED ISSUE #441: Collision filtering in View Data
**Description**
One of the core pieces of CRASH functionality is filtering collisions; MOVE currently has no equivalent feature.

This task adds an initial version of that feature, using a shortlist of filters gathered through user research.  See #408 for a similar task in our previous request management work - the general steps are the same, except that there are no top-level chips here.

**Acceptance Criteria**
- [x] in `viewData` store module, rename `filters` to `filtersStudy` and update getters, mutators, actions, and callsites accordingly;
- [x] add new store functionality for `filtersCollision`;
- [x] add filter selectors to filter dialog;
- [x] link filter dialog to `filtersCollision` in this store via `v-model`;
- [x] show filter chips next to filter button;
- [x] remove filters on clicking associated filter chip;
- [x] add REST API endpoint to fetch all collisions at a location;
- [x] actually filter collisions based on active filters;
- [x] investigate visual prominence on map (e.g. opacity) to draw visual focus to matching collisions in View Data mode.

**Additional Notes**
According to our database, some locations in the City experience roughly 1000 collisions per year.  This is _extremely_ rare (IIRC it's limited to high-volume expressway ramps or something like that), plus we're still in an early prototyping phase, so we can likely get away with loading all relevant collisions in the frontend and filtering there.

candu:
# Issue Addressed
This PR closes #441 .

# Description
We wire up frontend filters and filter chips to a backend implementation of collision filtering.  In the process, we clean up View Data for consistency between collision and count data presentation: the "total" is now the unfiltered total in both cases, while the numbers in the data table represent filtered summary statistics.

As part of that, we also rename some parameters in `CountController` for consistency with their counterparts elsewhere (`CollisionController`, `CollisionDAO`, `CountDAO`), to make this a bit more readable in the future.

# Tests
Ran `npm run test:test-db` to test database query changes.

Ran `npm run test:test-api` to test REST API changes - there are currently a couple of failures relating to hardcoded intersection `13446886`, which is not included in the development dataset.

Ran frontend / backend and manually tested collision and count filtering.


lib/api/WebApi.js
```diff
@@ -36,8 +37,22 @@ async function getAuth() {
   return AuthState.read.validateAsync(auth);
 }
 
-async function getCollisionPopupDetails(id) {
-  return apiClient.fetch(`/collisions/${id}/popupDetails`);
+async function getCollisionByCollisionId(collisionId) {
+  const collision = await apiClient.fetch(`/collisions/${collisionId}`);
```
candu, line 41: can we add some more tests on `CollisionDAO`?

lib/controller/CountController.js
```diff
@@ -67,18 +67,18 @@ CountController.push({
     const {
       centrelineId,
       centrelineType,
-      dayOfWeek: daysOfWeek,
-      end,
+      daysOfWeek,
```
candu, line 70: `CountDAO` is a bit odd - `byCentreline` expects to work on a single study type, while `byCentrelineSummary` accepts `studyTypes` as a filter (and looks at all study types if not provided).  in addition, `byCentreline` offers pagination.

in `CollisionDAO`, by contrast, both corresponding methods accept the same set of filters, while `byCentreline` does not offer pagination.

(IMHO this is fine as is, but we should probably document it eventually.)

lib/time/TimeFormatters.js
```diff
@@ -143,7 +143,10 @@ class TimeFormatters {
    * @returns {string} the formatted time range, using time of day for both range endpoints
    */
   static formatRangeTimeOfDay({ start, end }) {
-    const startHuman = TimeFormatters.formatTimeOfDay(start);
+    let startHuman = TimeFormatters.formatTimeOfDay(start);
+    if (startHuman === '24:00') {
```
candu, line 147: add a test case for this

candu:
fixed centreline ID thing in #454 , merging this one now.

PR #460, opened 2020-05-19
Collision report stubs with summary header

### LINKED ISSUE #442: Collision reports: drawer layout
**Description**
Another part of CRASH that's missing from MOVE right now is reporting: users of CRASH can view several reports on a filtered set of collisions.

This task adds that to MOVE.

**Acceptance Criteria**
- [x] top bar: "View Data", location description, filter chips;
- [x] tab bar for report types;
- [x] hook tab bar up to separate stub components for different reports, each tied to different routes;
- [x] add report classes, etc. in MOVE Reporter service for these report types;
- [x] in collision report base class, include header listing collision summary stats;
- [x] stub for Download button at top-right, similar to FLOW reports.

candu:
# Issue Addressed
This PR closes #442 .

# Description
We implement stub versions of `ReportCollisionDirectory` and `ReportCollisionTabulation`, which for now just show a header with the same summary statistics as the View Data page.

These implementations share base class `ReportBaseCrash`, which uses the collision filters as report parameters to fetch collisions via `CollisionDAO.byCentreline`.

# Tests
Ran the frontend, backend, and reporter, and tested manually in browser; `npm run test:test-db` to test database / DAO changes.

### DIFF
--- a/lib/model/helpers/CentrelineFeature.js
+++ b/lib/model/helpers/CentrelineFeature.js
@@ -1,0 +1,10 @@
+import { CentrelineType } from '@/lib/Constants';
+import Joi from '@/lib/model/Joi';
+
+export default {
+  centrelineId: Joi.number().integer().positive().required(),
+  centrelineType: Joi.number().valid(
+    CentrelineType.SEGMENT,
+    CentrelineType.INTERSECTION,
+  ).required(),
+};
--- a/lib/reports/ReportCollisionDirectory.js
+++ b/lib/reports/ReportCollisionDirectory.js
@@ -7,16 +7,19 @@ class ReportCollisionDirectory extends ReportBaseCrash {
     return ReportType.COLLISION_DIRECTORY;
   }
 
-  transformData(/* location, collisions, filters */) {
+  transformData(location, rawData) {
+    return rawData;
+  }
+
+  generateCsv(/* location, { collisions } */) {
     // TODO: implement this
   }
 
-  generateCsv(/* location, filteredCollisions */) {
-    // TODO: implement this
-  }
-
-  generateLayoutContent(/* location, filteredCollisions */) {
-    // TODO: implement this
+  generateLayoutContent(location, { collisionSummary }) {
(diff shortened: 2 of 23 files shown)


lib/model/helpers/CentrelineFeature.js
```diff
@@ -0,0 +1,10 @@
+import { CentrelineType } from '@/lib/Constants';
+import Joi from '@/lib/model/Joi';
+
+export default {
```
candu, line 4: one issue here is that we're conflating models for handling database persistence (e.g. `StudyRequest`) with other validation schemas, such as for endpoint query / response types.  we may need a separate space for the latter (e.g. `@/lib/schemas`, or some additional thought into how to bring these two use cases for `Joi` validation together.

lib/reports/ReportCollisionDirectory.js
```diff
@@ -7,16 +7,19 @@ class ReportCollisionDirectory extends ReportBaseCrash {
     return ReportType.COLLISION_DIRECTORY;
   }
 
-  transformData(/* location, collisions, filters */) {
-    // TODO: implement this
+  transformData(location, rawData) {
+    return rawData;
```
candu, line 11: why does this `transformData()` method get to share references with the result of `ReportBaseCrash#fetchRawData`?  we should either change the documentation on `ReportBase#fetchRawData`, or we should make this deep-clone `rawData`.

lib/model/helpers/CentrelineFeature.js
candu: the main difference between this and `GeoJson.Point` is that the latter is self-contained within a `Joi.object()` declaration, and is intended for use as the schema for individual fields within larger objects.  By contrast, this schema is more of a "mixin": it contains two fields, which are sometimes used standalone (e.g. in various endpoint query schemas) and sometimes used alongside other fields (e.g. in `StudyRequest`).

let's leave it for now, but we might (as mentioned above) eventually want to move it.

lib/reports/ReportCollisionDirectory.js
candu: (actually, this will probably change as we move towards building the reports, so let's leave it for now)

PR #476, opened 2020-06-02
Load testing harness, initial runs

### LINKED ISSUE #474: Load testing for MOVE REST API
**Description**
To understand the impact of upcoming multi-location features, and to prepare for eventual production release, we need to conduct load testing.

This task aims to build a load testing harness that we can use to prep for multi-location features and later expand upon for pre-release testing.

**Acceptance Criteria**
- [x] select a tool to conduct HTTP load testing against the API;
- [x] build lists of URLs to call for testing;
- [x] build a script that runs tests of increasing intensity against each list of URLs;
- [x] successfully run a load test from AWS;
- [x] analyze results to understand raw performance.

**Additional Notes**
`httpload` and `wrk2` both look like promising tools here - need to investigate further.

For the AWS tests, the goal is to learn how much traffic MOVE can handle on different REST API endpoints, with the intent of extrapolating from there to understand what multi-location load could look like.

candu:
# Issue Addressed
This PR closes #474 .

# Description
We add `tests/load`, which contains a testing harness that uses [`wrk2`](https://github.com/giltene/wrk2) to run a series of load tests.  These load tests use a custom Lua script, `path_list.lua`, that loads path lists as built by `build_path_lists.sh` and randomly selects paths from those lists for each request.

From the output, we can see:

- maximum requests-per-second (RPS) for various endpoints;
- latency / RPS curves for those same endpoints.

# Tests
Ran load testing, observed output.  Main takeaways (copied from elsewhere):

- report generation is likely to be an issue in a multi-location setup.  if we want to generate reports for more than a couple of locations at once, that will likely have to take place in a background queue, or perhaps ahead of time somehow (though the filtered nature of CRASH reports means that pre-building is probably only feasible for FLOW reports.)
- we're also likely to need more instances of the reporter service - I think that can be managed via nginx without asking for anything else from Cloud Services, although I'll need to figure out what the limiting resource is to determine if adding more on the same machine as our web service would have adverse effects on the latter.
- View Data, on the other hand, is much more likely to be able to handle multi-location load.  the queries that load summary stats there are much faster, and can handle 2-3 orders of magnitude more requests per second than the reporting endpoints.
- that said: I don't yet have a good sense for how people might use multi-location features in practice, which makes it very hard to estimate what we could get away with in View Data or how many additional reporter instances we might need.  without any additional performance work, 10 locations would probably work, and 1000 locations probably wouldn't, and 1-2 more reporter instances should easily handle longitudinal test load, but I'd need more info from both qualitative and quantitative standpoints to tighten those bounds and understand what production-readiness might look like here.


tests/load/build_path_lists.sh
```diff
@@ -0,0 +1,47 @@
...
+grep -E "^5" features/counts.txt | sed -r 's/^(.*)\t(.*)$/\/reporter\/reports?type=COUNT_SUMMARY_TURNING_MOVEMENT_DETAILED\&id=\1%2F\2\&format=WEB/' > paths/getReportWeb_COUNT_SUMMARY_TURNING_MOVEMENT_DETAILED.txt
+grep -E "^5" features/counts.txt | sed -r 's/^(.*)\t(.*)$/\/reporter\/reports?type=INTERSECTION_SUMMARY\&id=\1%2F\2\&format=PDF/' > paths/getReport_INTERSECTION_SUMMARY.txt
+grep -E "^5" features/counts.txt | sed -r 's/^(.*)\t(.*)$/\/reporter\/reports?type=INTERSECTION_SUMMARY\&id=\1%2F\2\&format=WEB/' > paths/getReportWeb_INTERSECTION_SUMMARY.txt
+grep -E "^4" features/counts.txt | sed -r 's/^(.*)\t(.*)$/\/reporter\/reports?type=SPEED_PERCENTILE\&id=\1%2F\2\&format=PDF/' > paths/getReport_SPEED_PERCENTILE.txt
+grep -E "^4" features/counts.txt | sed -r 's/^(.*)\t(.*)$/\/reporter\/reports?type=SPEED_PERCENTILE\&id=\1%2F\2\&format=WEB/' > paths/getReportWeb_SPEED_PERCENTILE.txt
+grep -E "^5" features/counts.txt | sed -r 's/^(.*)\t(.*)$/\/reporter\/reports?type=WARRANT_TRAFFIC_SIGNAL_CONTROL\&id=\1%2F\2\&format=PDF/' > paths/getReport_WARRANT_TRAFFIC_SIGNAL_CONTROL.txt
```
candu, line 46: these endpoints probably need additional parameters - the `status` error rates are very high.  (this also means possibly redoing this particular test.)

PR #481, opened 2020-06-12
Fully integrate multi-day, multi-direction studies into MOVE reporting

### LINKED ISSUE #201: Multi-day counts show up as separate single-day counts
**What Happened**
When viewing any location with a multi-day ATR count, the count shows up as separate single-day counts, one per day.

**What Should Happen**
ATR counts on consecutive days with the same `arterycode` should be treated as a single multi-count.

**To Reproduce**
Steps to reproduce the bug:
1. Go to "View Map";
2. Select any location for which there is a multi-day ATR count;
3. In the counts table, hover over the date dropdown.

At this point, you'll see separate counts listed for each day of the multi-day ATR count.

**Additional Notes**
It's unclear what to do for RESCU or other permanent station counts, where locations can have over 10k counts.  We may need to continue treating those as individual single-day counts.

This may involve some ETL pipeline work as well, e.g. if we want to provide a `MATERIALIZED VIEW` that solves this problem.  In that case, we'd also need to modify existing pipelines to `TRUNCATE`.

candu:
# Issue Addressed
This PR closes #201 🎉 

# Description
We integrate multi-day, multi-direction studies into MOVE reporting by updating each FLOW report type, one by one.

# Tests
Updated unit tests for individual report types and ran those.  Tested these from the frontend, running all three components of MOVE and clicking around a variety of locations to view different report types.

### DIFF
--- a/tests/jest/unit/reports/ReportCountSummary24h.spec.js
+++ b/tests/jest/unit/reports/ReportCountSummary24h.spec.js
@@ -1,17 +1,19 @@
 /* eslint-disable camelcase */
-import path from 'path';
+// import path from 'path';
 
-import { CardinalDirection, StudyType } from '@/lib/Constants';
+// import { CardinalDirection, StudyType } from '@/lib/Constants';
 import ReportCountSummary24h from '@/lib/reports/ReportCountSummary24h';
-import { loadJsonSync } from '@/lib/test/TestDataLoader';
+// import { loadJsonSync } from '@/lib/test/TestDataLoader';
 import DateTime from '@/lib/time/DateTime';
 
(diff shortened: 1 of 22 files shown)


tests/jest/unit/reports/ReportCountSummary24h.spec.js
```diff
@@ -75,6 +77,7 @@ test('ReportCountSummary24h.timeRange', () => {
   });
 });
 
+/*
```
candu, line 80: can we modify this test in the same way as other tests?  we shouldn't just stop testing this one report type.

PR #500, opened 2020-06-23
Add arterycode, signal number support in location search

### LINKED ISSUE #401: Add traffic signal IDs, artery codes to new location search
**Description**
Now that we finally have a more flexible location search implementation, we can integrate previous work by @chelsrosic on adding more sources to the search bar.

This task does just that, integrating her work on traffic signals and pedestrian crossovers as well as artery / station codes.

**Acceptance Criteria**
- [x] break out `CentrelineDAO.intersectionSuggestions` into a separate DAO (e.g. `LocationSearchDAO`);
- [x] implement `LocationSearchDAO.signalSuggestions`:
  - [x] create view in `location_search` schema on `gis.traffic_signals`;
  - [x] given a PX ID, return matching features from `location_search.traffic_signals`;
- [x] implement `LocationSearchDAO.arterySuggestions`, which searches artery codes;
- [x] update `LocationController.getLocationSuggestions` to detect `px:` and other prefix terms, and call the appropriate method of `LocationSearchDAO`;
- [ ] (optional) figure out how to handle compound queries (see below).

**Additional Notes**
This continues the approach @chelsrosic worked on, wherein we use prefix terms like `px:`, `artery:`, etc. to trigger querying specific data sources.  For example, `px:1234` (or: `tsc:1234`) would search for all signals whose ID starts with `1234`, preferring an exact match.

`SignalDAO` contains SQL queries to search traffic signals and pedestrian crossovers, while upcoming work on #358 will put artery / station codes in a normalized format conducive to search integration.

The target prefix queries here:

- signal suggestions: `px:`, `tsc:`, and `signal:` all work;
- artery codes: `artery:` and `station:` both work.

The idea is that we start with a "best guess" set of aliases, and slowly change them in response to user feedback.  This means we design for this flexibility up front!

We'll have to figure out how to handle queries like `px:123 Baldwin`.  We have a few options:

- ignore the whole query as malformed;
- ignore the `Baldwin` part and just search for `px:123`;
- treat this as an OR query, combining results for `px:123` and results for `Baldwin`;
- treat this as an AND query, returning only those results for `px:123` that textually match `Baldwin`.

In the same vein: what about queries like `px:123 station:4567`, or even `px:123 Baldwin station:4567`?

In search systems, one common approach is to break the query into individual subqueries, execute all subqueries, then combine the results using set union / intersection.  The result sets for the subqueries are often referred to as [postings lists](https://nlp.stanford.edu/IR-book/html/htmledition/an-example-information-retrieval-problem-1.html) - effectively, this is a list of `(id, score)` pairs, where `id` uniquely identifies features and `score` is a ranking _within that postings list_.  This allows both set operations on `id` and ranking on `score` once postings lists are combined.

For example, given the query `px:123 Baldwin station:4567`:

- break it into subqueries: `['px:123', 'Baldwin', 'station:4567']`;
- execute `px:123`, `Baldwin`, `station:4567` in parallel, getting their postings lists;
- combine postings lists (e.g. using set union);
- rank the combined result set (e.g. give precedence to matches on earlier queries, then use internal ranking of each postings list).

candu:
# Issue Addressed
This PR closes #401 .

# Description
We add methods for arterycode and signal number search to `LocationSearchDAO`, plus a new `LocationSearchDAO.getSuggestions` method that stitches these together with textual intersection search.

In the process, we move centreline feature normalization functions out of `CentrelineDAO` and into a common `NormalizeUtils` module, where they can be used both by `CentrelineDAO` and `LocationSearchDAO`.

# Tests
Tested in frontend, and via the REST API.


lib/db/CentrelineDAO.js
```diff
@@ -125,138 +83,6 @@ SELECT DISTINCT ON (int_id)
     return rows.map(intersectionToFeature);
   }
 
-  /**
```
candu, line 128: update these DAO tests, please

lib/db/LocationSearchDAO.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  static async getSuggestions(query, limit) {
+    const queryNormalized = query.trim();
+
+    const match = REGEX_SPECIAL_TERM.exec(queryNormalized);
+    if (match !== null) {
+      const value = parseInt(match[2], 10);
+      if (Number.isNaN(value)) {
+        return [];
```
candu, line 209: eventually it would be nice to have error states on this, but we're not quite there yet

PR #503, opened 2020-06-26
CompositeId: identifier codec for location feature sets

### LINKED ISSUE #502: Create composite ID library to handle multi-location selection IDs
**Description**
One crucial piece of our multi-location features is _identification_: selections need _identifiers_ that uniquely and unambiguously define the features they contain.  These allow us to pass those selections between various components of MOVE:

- from MOVE itself to emails and copied permalinks;
- from MOVE's frontend to REST API endpoints in the MOVE web application;
- from MOVE's frontend to MOVE Reporter, to generate a series of reports;
- (possibly) from various backend components to the database.

**Acceptance Criteria**
- [x] create `CompositeId` class and test suite;
- [x] encode a list of feature identifiers into a shortened URL-friendly string;
- [x] decode that string back into feature identifiers;
- [x] add robust tests, including fuzz tests, for all encode / decode operations.

candu:
# Issue Addressed
This PR closes #502 .

# Description
We introduce `CompositeId`, a "codec" (encoder / decoder) for location feature sets.  It uses `BitStream` to build up a compact binary representation of a set of location features.  `BitStream#toString` serializes that binary representation to a URL-safe base64 string, while `BitStream.fromString` deserializes that string back to the binary representation.  `CompositeId` wraps this serialization / deserialization into its `encode` and `decode` methods.

`decode` and `fromString` perform extra validation; the corresponding unit tests include tests for these error conditions.  We also provide a full set of test cases, including fuzz tests that try encode / decode cycles with randomly-generated feature sets.

# Tests
See above description: `npx jest BitStream.spec.js` and `npx jest CompositeId.spec.js` test these two libraries.  Pre-commit checks pass.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9464,9 +9464,9 @@
       }
     },
     "base64-js": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.3.0.tgz",
-      "integrity": "sha512-ccav/yGvoa80BQDljCxsmmQ3Xvx60/UpBIij5QN21W3wBi/hhIC9OoO+KLpu9IJTS9j4DRVJ3aDDF9cMSoa2lw=="
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.3.1.tgz",
+      "integrity": "sha512-mLQ4i2QO1ytvGWFWmcngKO//JXAQueZvwEKtjgQFM4jIK0kU+ytMfplL8j+n5mspOfjHwoAg+9yhb7BwAHm36g=="
     },
     "base64url": {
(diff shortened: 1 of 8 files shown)


package-lock.json
```diff
@@ -9464,9 +9464,9 @@
       }
     },
     "base64-js": {
-      "version": "1.3.0",
-      "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.3.0.tgz",
-      "integrity": "sha512-ccav/yGvoa80BQDljCxsmmQ3Xvx60/UpBIij5QN21W3wBi/hhIC9OoO+KLpu9IJTS9j4DRVJ3aDDF9cMSoa2lw=="
+      "version": "1.3.1",
```
candu, line 9467: assuming this is a dependency of some other library...

Issue #507, opened 2020-06-26
Multi-location: stub out location selector box


candu:
**Description**
Multi-location designs include a selector box at top-left for selecting multiple locations.  This task refactors frontend code to stub that box out for now.

**Acceptance Criteria**
- [x] decouple `FcSearchBarLocation` from `vuex` store, and instead build it using `FcMixinVModelProxy`;
- [x] in single-location mode, wrap `FcSearchBarLocation` in a component that handles syncing to `vuex` store;
- [x] swap single-location pieces of state in `vuex` store out for multi-location selections;
- [x] stub out Edit mode layout for multi-location selector, and add at top-left of map;
- [x] "Cancel" button switches back to single-location mode;
- [x] reuse `FcSearchBarLocation` within multi-location selector, and adjust visual style to create "stacking" effect.

For now, the "include..." checkbox and "Done" button should do nothing.

candu:
Closed by #521 (but forgot to tag this issue there.)

PR #518, opened 2020-06-29
Multi-location support for collision, study, and study request APIs

candu:
# Issue Addressed
This PR makes progress on #507 .

# Description
We change collision, study, and study request APIs that fetch data by centreline feature to support multi-location feature selections, backed by `CompositeId` and the new `Joi.compositeId()` validator.

We then change `WebApi` to wrap existing single-location calls into this new multi-location setup, so that we can shift towards multi-location backend calls without immediately changing frontend callsites.

# Tests
Updated REST API and DAO tests.  Loaded View Data in the frontend to verify that it still works properly.


candu:
Rubber-stamping this one for now to move forward.

PR #521, opened 2020-06-30
Multi-location state in vuex store

### LINKED ISSUE #407: Request Management: Track Requests search
**Description**
The designs referred to in #405 include a search bar, which is used to perform full-text search on the Track Requests table.

This task implements this search bar.

**Acceptance Criteria**
- [x] refactor search bar into a separate component (e.g. `FcSearchBarRequests`);
- [x] add local `data` fields in `FcRequestsTrack`, and link them via `v-model` to that component;
- [ ] debounce setter in search bar component (e.g. 250 ms after last input);
- [x] implement column-specific searches (see below):
  - [x] Location
  - [x] Type
  - [x] Requester
  - [x] Date Created
  - [x] Assigned To
- [x] implement "All Columns" search (see below).

**Additional Notes**
The different columns introduce their own challenges, as explored below.

"Location" is the location description, as would be returned from `CentrelineDAO`.  This means that we _could_ use something similar to `FcSearchBarLocation`, with the autocomplete and everything.  OTOH, a simpler prefix-match approach might be good enough here - we should test that first, and only bring in something more complicated if needed.

"Type" is the study type.  After #403 , each request only has one study, which simplifies this a bit.  Each `StudyType` (see `@/lib/Constants`) has both a `name` and a `label`, but the `name` is never shown to the user.  However, terms like `ATR` and `TMC` are commonly used to refer to these queries - we may need an additional field in `StudyType` that lists search keywords for each type.  There's also the consideration of how / whether these link to study type filters.  (IMHO, they probably shouldn't.)

"Requester" is the most straightforward: we can do substring match on the query.  We should probably ignore the `CHILD\` part.

"Date Created" introduces the challenge of searching on dates based on a text query.  To start with, let's allow substring matches against the formatted date.  (We can expand that as testing results demand.)

"Assigned To" only has three valid values: OTI, Staff, and None.  This might make it a better candidate for filters.

For "All Columns", we can run the query on each individual column, then `OR` the results.

There's also a big question around where search is performed: do we filter in the database, in the web application backend, or in the frontend?  Some columns (e.g. "Type") don't really lend themselves to pure SQL-based search, as they depend on metadata in the `StudyType` enum.  However, if we push filtering into the database, we can more easily paginate / stream results later.

As mentioned in #408 , the wisest course of action for now is to fetch everything and search in frontend.  This allows for quicker implementation and iteration, which is important until we surface the finer details through usability testing.

candu:
# Issue Addressed
This PR closes #407 .

# Description
We change the way location selections are stored in `vuex` from a single location to an array of locations.  We then refactor a whole slew of callsites to use this new multi-location array: reading it from `state.locations`, writing it via the `setLocations` mutation, encoding it into a `CompositeId` for navigation between frontend routes, and passing that encoded `CompositeId` to REST API endpoints for multi-location data lookup in the backend.

The end result is that we can now largely roll the multi-location ball downhill, building UI changes into the framework already built up here and in previous PRs.  (There is still a substantial amount of work on those UI changes.)

# Tests
Tested several user flows in the frontend, including reporting flows.

### DIFF
--- a/lib/api/WebApi.js
+++ b/lib/api/WebApi.js
@@ -47,8 +47,8 @@ async function getCollisionByCollisionId(collisionId) {
   return CollisionEvent.read.validateAsync(collision);
 }
 
-async function getCollisionsByCentreline(feature, filters) {
-  const s1 = CompositeId.encode([feature]);
+async function getCollisionsByCentreline(features, filters) {
+  const s1 = CompositeId.encode(features);
   const data = { s1, ...filters };
   const options = { data };
   const collisions = await apiClient.fetch('/collisions/byCentreline', options);
@@ -56,27 +56,27 @@ async function getCollisionsByCentreline(feature, filters) {
   return collisionsSchema.validateAsync(collisions);
 }
 
-async function getCollisionsByCentrelineSummary(feature, filters) {
-  const s1 = CompositeId.encode([feature]);
+async function getCollisionsByCentrelineSummary(features, filters) {
+  const s1 = CompositeId.encode(features);
   const data = { s1, ...filters };
   const options = { data };
   return apiClient.fetch('/collisions/byCentreline/summary', options);
 }
 
-async function getCollisionsByCentrelineTotal(feature) {
-  const s1 = CompositeId.encode([feature]);
+async function getCollisionsByCentrelineTotal(features) {
+  const s1 = CompositeId.encode(features);
   const data = { s1 };
   const options = { data };
   const { total } = await apiClient.fetch('/collisions/byCentreline/total', options);
   return total;
 }
 
-async function getLocationsByFeature(centrelineTypesAndIds) {
-  if (centrelineTypesAndIds.length === 0) {
+async function getLocationsByFeature(features) {
+  if (features.length === 0) {
     return new Map();
   }
-  const centrelineIds = centrelineTypesAndIds.map(({ centrelineId: id }) => id);
-  const centrelineTypes = centrelineTypesAndIds.map(({ centrelineType: type }) => type);
+  const centrelineIds = features.map(({ centrelineId: id }) => id);
+  const centrelineTypes = features.map(({ centrelineType: type }) => type);
   const options = {
     data: {
       centrelineId: centrelineIds,
@@ -164,8 +164,8 @@ async function getReportWeb(type, id, reportParameters) {
   };
 }
 
-async function getStudiesByCentreline(feature, studyType, filters, pagination) {
-  const s1 = CompositeId.encode([feature]);
+async function getStudiesByCentreline(features, studyType, filters, pagination) {
+  const s1 = CompositeId.encode(features);
   const data = { s1, ...filters, ...pagination };
   const options = { data };
   const studies = await apiClient.fetch('/studies/byCentreline', options);
@@ -173,8 +173,8 @@ async function getStudiesByCentreline(feature, studyType, filters, pagination) {
   return studiesSchema.validateAsync(studies);
 }
 
-async function getStudiesByCentrelineSummary(feature, filters) {
-  const s1 = CompositeId.encode([feature]);
+async function getStudiesByCentrelineSummary(features, filters) {
+  const s1 = CompositeId.encode(features);
   const data = { s1, ...filters };
   const options = { data };
   const studySummary = await apiClient.fetch('/studies/byCentreline/summary', options);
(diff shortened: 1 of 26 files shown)


lib/api/WebApi.js
```diff
@@ -47,36 +47,36 @@ async function getCollisionByCollisionId(collisionId) {
...
 
-async function getLocationsByFeature(centrelineTypesAndIds) {
-  if (centrelineTypesAndIds.length === 0) {
+async function getLocationsByFeature(features) {
+  if (features.length === 0) {
     return new Map();
   }
-  const centrelineIds = centrelineTypesAndIds.map(({ centrelineId: id }) => id);
-  const centrelineTypes = centrelineTypesAndIds.map(({ centrelineType: type }) => type);
+  const centrelineIds = features.map(({ centrelineId: id }) => id);
```
candu, line 78: why don't we just change this endpoint so that it resolves a `CompositeId` to its features?

lib/reports/ReportBaseCrash.js
```diff
@@ -19,58 +21,35 @@ class ReportBaseCrash extends ReportBase {
...
-
+  async parseId(s1) {
     try {
-      const location = await CentrelineDAO.byIdAndType(centrelineId, centrelineType);
-      if (location === null) {
-        throw new InvalidReportIdError(rawId);
-      }
-      return location;
+      const features = CompositeId.decode(s1);
+      const locations = await CentrelineDAO.byIdsAndTypes(features);
```
candu, line 33: ideally we would rename this to `CentrelineDAO.byFeatures` for readability

lib/reports/ReportBaseCrash.js
```diff
@@ -19,58 +21,35 @@ class ReportBaseCrash extends ReportBase {
...
+      return Array.from(locations.values());
     } catch (err) {
       if (err instanceof InvalidCentrelineTypeError) {
-        throw new InvalidReportIdError(rawId);
+        throw new InvalidReportIdError(s1);
```
candu, line 37: don't rethrow this as `InvalidReportIdError` - if the decoding works, it should guarantee that the resulting features have valid `centrelineType` values.

web/components/FcDrawerViewData.vue
```diff
@@ -419,11 +401,11 @@ export default {
...
       this.studyTotal = studyTotal;
 
-      if (this.location === null
-          || location.centrelineId !== this.location.centrelineId
-          || location.centrelineType !== this.location.centrelineType
-          || location.description !== this.location.description) {
-        this.setLocation(location);
+      const { s1 } = this;
+      if (s1 !== s1Next) {
+        const locationMap = await getLocationsByFeature(features);
```
candu, line 406: this suggests that `getLocationsByFeature` should *not* return a `Map`, but instead an `Array` - we can always build a map out of an `Array`, e.g. as

```
const locationsMap = new Map(
  locations.map(location => [centrelineKey(location), location]),
);

File: web/components/FcPaneMapPopup.vue [hunk 5]
```diff
@@ -383,18 +384,20 @@ export default {
...
+      this.setLocations([location]);
     },
     actionViewData() {
       if (this.$route.name === 'viewDataAtLocation') {
         this.setDrawerOpen(true);
       }
 
       // open the view data window
-      const { centrelineId, centrelineType } = this.feature.properties;
+      const features = [this.feature.properties];
```
candu, line 395: let's be more explicit here:

```
const { centrelineId, centrelineType } = this.feature.properties;
const feature = { centrelineId, centrelineType };
const features = [feature];
```

web/components/FcPaneMapPopup.vue
```diff
@@ -383,18 +384,20 @@ export default {
...
+      const s1 = CompositeId.encode(features);
+      console.log(s1);
```
candu, line 397: remove `console.log` statement

web/store/LoginState.js
```diff
@@ -9,8 +9,8 @@ function restoreLoginState(next) {
   }
   window.sessionStorage.removeItem(STORAGE_KEY_LOGIN_STATE);
   try {
-    const { location, name, params } = JSON.parse(loginState);
-    store.commit('setLocation', location);
+    const { locations, name, params } = JSON.parse(loginState);
```
candu, line 12: test this interaction, please

web/store/index.js
```diff
@@ -70,8 +74,17 @@ export default new Vuex.Store({
     },
     // LOCATION
     locationFeatureType(state) {
-      const { location } = state;
-      return getLocationFeatureType(location);
+      const { locations } = state;
```
candu, line 77: we have a couple of places (e.g. map, Request Study) where we offer a computed property `location`, which is really just the first location (or `null` if the selection is empty) - this allows us to migrate gradually without breaking existing flows.

can we move that computed property in here as a getter, and add a TODO on it to remove once we're much closer to finishing?

PR #522, opened 2020-07-06
Gh520 bitstream limits

### LINKED ISSUE #520: Limit size of BitStream / CompositeId
**What Happened**
There is currently no limit on the declared bit length of a `BitStream` or `CompositeId`.  This means that a user could pass `?s1=s1:100000000:` to any endpoint that takes an `s1` encoded `CompositeId` parameter, and right now it would allocate a 500 MB `Uint8Array` before even noticing the length mismatch.  By crafting a `CompositeId` with a high enough declared bit length, this can cause a denial-of-service via OOM crash.

**What Should Happen**
There should be a limit _somewhere_ on the declared length of `CompositeId` values passed to our APIs, and this limit should be enforced as soon as possible - and definitely before allocating the decoding buffer.

**Additional Notes**
We may also want to add a rule in our `Joi.compositeId()` wrapper, so that this can be verified at time of API call.

As per [this SO question](https://stackoverflow.com/questions/32763165/node-js-http-get-url-length-limitation), `node` enforces a limit of 80kB for headers + URL.  That corresponds to ~16k features, which is much larger than any practical multi-location selection limit.

One possibility is to change the `BitStream` encoding itself: we could drop the `len` section and add a character at the end of the `b64` section whose purpose is to encode `bitLength % 6`.  We could use any six base 64 characters, or perhaps the "marks" from [RFC 2396](https://www.ietf.org/rfc/rfc2396.txt) (e.g. `.!~*()`) if we need to distinguish this character from `b64` values.  We could also replace the `len` section with a `mod` section for this purpose, such that the bit length would become `b64.length * 6 + decodeMod(mod)`.  By doing this, we get rid of the potential to craft a malicious `len` value - we would still need an explicit limit, but that limit becomes easier to impose.

candu:
# Issue Addressed
This PR closes #520 .

# Description
We change the format of `BitStream` to `${mod}${b64}`, where `${mod}` is a single base-64 character representing `bitLength % 6` and `${b64}` contains base-64-encoded bits as before.  We also impose a limit, `CompositeId.MAX_LENGTH`, on the size of a `CompositeId`-encoded string.

# Tests
Ran `BitStream` and `CompositeId` tests via `jest`.

### DIFF
--- a/lib/io/CompositeId.js
+++ b/lib/io/CompositeId.js
@@ -65,6 +65,9 @@ class CompositeId {
 
   static encode(features) {
     const n = features.length;
+    if (n > CompositeId.MAX_FEATURES) {
+      throw new InvalidCompositeIdError('maximum number of features exceeded');
+    }
     const numBytes = Math.ceil(n * CompositeId.BITS_PER_FEATURE / 8);
     const bytes = new Uint8Array(numBytes);
     const bitStream = new BitStream(bytes);
--- a/lib/io/CompositeId.js
+++ b/lib/io/CompositeId.js
@@ -20,6 +20,9 @@ import BitStream from '@/lib/io/BitStream';
  */
 class CompositeId {
   static decode(compositeId) {
+    if (compositeId.length > CompositeId.MAX_LENGTH) {
+      throw new InvalidCompositeIdError('maximum length exceeded');
+    }
     const typeSeparatorIndex = compositeId.indexOf(':');
     if (typeSeparatorIndex === -1) {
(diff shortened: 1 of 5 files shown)


lib/io/CompositeId.js
```diff
@@ -20,6 +20,9 @@ import BitStream from '@/lib/io/BitStream';
  */
 class CompositeId {
   static decode(compositeId) {
+    if (compositeId.length > CompositeId.MAX_LENGTH) {
```
candu, line 23: can we also check against `CompositeId.MAX_FEATURES` in `encode()`?

PR #524, opened 2020-07-09
Multi-location selection in map

### LINKED ISSUE #508: Multi-location: Edit mode for location selector box
**Description**
Multi-location designs include a selector box at top-left for selecting multiple locations.  This task implements the Edit mode of that box, integrates the set of selected locations with the map, and properly handles transitions into multi-location and back out again.

**Acceptance Criteria**
- [x] implement adding a new location;
- [x] implement 5 location limit by showing message below when limit is reached;
- [x] implement "X" buttons at right;
- [x] implement editing previously selected locations;
- [x] update markers on map to reflect selected locations;
- [x] when clicking on map, edit active selected location (or add to list if needed).

candu:
# Issue Addressed
This PR closes #508 .

# Description
We update `FcPaneMap` to handle multi-location selections by adding new `locations` and `active` centreline layers.  `locations` layers are backed by a GeoJSON source that is updated to match the `locations` state from `vuex`, while `active` layers use Mapbox GL [filter expressions](https://docs.mapbox.com/mapbox-gl-js/example/query-similar-features/) to reflect hover / click interactions with individual map features.

As part of this, we also sever the connection between the selected location and hovered / selected features on the map - these must now be set separately.  We also refactor `easeToLocation` into the new `easeToLocations`, which builds a bounding box on multiple features before zooming.

Finally, we resolve an interaction bug with blur / focus states in `FcInputLocationSearch`.

# Tests
Quick manual testing in frontend.

### DIFF
--- a/lib/geo/GeoStyle.js
+++ b/lib/geo/GeoStyle.js
@@ -385,10 +385,10 @@ class GeoStyle {
     return STYLE_BASE_MAP_LIGHT;
   }
 
-  static getNonSymbolLayers({ aerial, dark, layers: { collisions, counts, volume } }) {
+  static getNonSymbolLayers({ aerial, dark, layers: { collisions, studies, volume } }) {
     if (aerial) {
       let paint = {};
-      if (collisions || counts || volume) {
+      if (collisions || studies || volume) {
         paint = {
           'raster-brightness-min': 0.2,
           'raster-opacity': 0.8,
--- a/lib/geo/GeoStyle.js
+++ b/lib/geo/GeoStyle.js
@@ -242,7 +242,6 @@ function injectSources(style) {
   addDynamicTileSource(style, 'studies:5', MapZoom.LEVEL_2, MapZoom.LEVEL_1);
   addDynamicTileSource(style, 'studies:10', MapZoom.LEVEL_2, MapZoom.LEVEL_1);
   addGeoJsonSource(style, 'locations');
-  addGeoJsonSource(style, 'active');
 }
 
 injectSources(STYLE_BASE_MAP_DARK);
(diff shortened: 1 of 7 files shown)


lib/geo/GeoStyle.js
```diff
@@ -436,6 +437,27 @@ class GeoStyle {
           ],
         },
       }),
+      {
+        id: 'locations-midblocks',
+        source: 'locations',
```
candu, line 442: IMHO we need to switch to a model where each `FcPaneMap` instance has its own `GeoStyle` instance.  Right now, changing the map options blows away the `locations` GeoJSON source data (tested this with both layer toggles and aerial mode) - I _suspect_ this is because we're rebuilding the style from the ground up each time we call `GeoStyle.get`.

web/components/FcPaneMap.vue
```diff
@@ -225,26 +270,16 @@ export default {
       'drawerOpen',
       'legendOptions',
       'locationMode',
+      'locations',
+      'locationsEdit',
```
candu, line 274: we don't use this anywhere right now!  in `locationsGeoJson`, we should be toggling between `this.locations` and `this.locationsEdit` depending on whether `this.locationMode === LocationMode.MULTI_EDIT`.

web/components/FcPaneMap.vue
```diff
@@ -345,43 +384,35 @@ export default {
...
+          if (type === 'Point') {
+            bounds.extend(coordinates);
+          } else if (type === 'LineString') {
+            coordinates.forEach((coordinatesPoint) => {
+              bounds.extend(coordinatesPoint);
+            });
+          }
         });
-      } else if (oldLocation === null) {
+        const cameraOptions = this.map.cameraForBounds(bounds, { padding: 40 });
```
candu, line 412: we should set `{ maxZoom: MapZoom.LEVEL_1.minzoom }` in the options here as well.  also separate camera options and add comment there too.

web/components/inputs/FcInputLocationSearch.vue
```diff
@@ -169,6 +166,29 @@ export default {
...
+       * handler is processed.
+       *
+       * By detecting this case here, we allow the state transition to
+       * `LocationSearchState.VALUE_SELECTED` within `actionSelect` to hide the dropdown instead,
+       * after the `@click` handler is processed.
+       */
+      const { relatedTarget: $relatedTarget } = e;
+      if ($relatedTarget !== null) {
+        const $list = $relatedTarget.closest('.fc-list-location-suggestions');
+        if (this.$refs.listLocationSuggestions !== undefined
```
candu, line 186: add comment here explaining need for `undefined` check, with [link to docs](https://vuejs.org/v2/guide/instance.html#Lifecycle-Diagram)

lib/geo/GeoStyle.js
candu: nvm, found a simpler solution here.

Issue #510, opened 2020-06-26
Multi-location: corridor interpolation in Edit mode


candu:
**Description**
In Edit mode, the user can choose whether they want to interpolate a corridor between their selected locations by checking (or unchecking) the "include..." checkbox.

This task implements corridor interpolation, and integrates it within the location selector box and map.

**Acceptance Criteria**
- [x] in `vuex` store, separate the location selector state from the location selection;
- [x] add DAO methods for routing:
  - [x] between intersections;
  - [x] between features;
  - [x] along a multi-waypoint corridor;
- [x] add REST API endpoint that takes a location selection and uses the DAO method to return the corresponding corridor;
- [x] in location selector box, add watchers on "include..." and the list of selected locations that:
  - [x] if "include..." checked, makes a request to the REST API endpoint, and sets the location selection to the response;
  - [x] otherwise, sets the location selection to the list of selected locations;
- [x] show corridor on map: square-hole pins for intersections, filled-in line for midblocks.

PR #526, opened 2020-07-13
Backend implementation of corridor interpolation 

candu:
# Issue Addressed
This PR makes significant progress on #510 .

# Description
We introduce `RoutingDAO` to deal with routing corridors to connect multi-location selections, using `pgrouting` (thanks @jovenc @chmnata and others for your excellent notes here from previous BDITTO work!)  We also add tests for routing methods in `RoutingDAO.spec.js`.

We then add a new endpoint, `LocationController.getLocationsByCorridor`, that accepts a `CentrelineSelection` and returns a `CentrelineLocation` array representing a corridor on the waypoints in that selection.

# Tests
Pre-commit tests, `npm run test:test-db`, `npm run test:test-api`.

### DIFF
--- a/lib/ArrayUtils.js
+++ b/lib/ArrayUtils.js
@@ -218,6 +218,36 @@ class ArrayUtils {
     });
     return groups;
   }
+
+  /**
+   * Similar to the UNIX utility `uniq` and lodash's `_.sortedUniq`, returns `xs` with
+   * consecutive similar values removed.  Two values `x1`, `x2` are similar iff
+   * `key(x1) === key(x2)`.
+   *
+   * Note that this does not handle non-consecutive similar values.
+   *
+   * @param {Array} xs
+   * @param {ComparisonKey} key
+   * @see https://lodash.com/docs/4.17.15#sortedUniq
+   */
+  static consecutiveUniqBy(xs, key) {
+    const n = xs.length;
+    if (n === 0) {
+      return [];
+    }
+    let x = xs[0];
+    const xsConsecutiveUniq = [x];
+    let kPrev = key(x);
+    for (let i = 1; i < n; i++) {
+      x = xs[i];
+      const k = key(x);
+      if (k !== kPrev) {
+        xsConsecutiveUniq.push(x);
+        kPrev = k;
+      }
+    }
+    return xsConsecutiveUniq;
+  }
 }
 
 export default ArrayUtils;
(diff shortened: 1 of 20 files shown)


lib/ArrayUtils.js
```diff
@@ -218,6 +218,36 @@ class ArrayUtils {
...
+   * consecutive similar values removed.  Two values `x1`, `x2` are similar iff
+   * `key(x1) === key(x2)`.
+   *
+   * Note that this does not handle non-consecutive similar values.
+   *
+   * @param {Array} xs
+   * @param {ComparisonKey} key
+   * @see https://lodash.com/docs/4.17.15#sortedUniq
+   */
+  static consecutiveUniqBy(xs, key) {
```
candu, line 233: I _think_ it's still the right call to avoid a dependency on `lodash`, if only to help keep build times and bundle sizes manageable - but we should maybe revisit that decision later on if we keep adding more methods to `ArrayUtils`.

lib/api/WebApi.js
```diff
@@ -87,6 +87,16 @@ async function getLocationByCentreline(feature) {
...
 }
 
+async function getLocationsByCorridor(features) {
+  if (features.length === 0) {
+    return [];
+  }
+  const s1 = CompositeId.encode(features);
+  const data = { s1 };
+  const options = { data };
+  return apiClient.fetch('/locations/byCorridor', options);
```
candu, line 97: now that we have response schemas on `LocationController`, can we enforce them in the client as well just like for other endpoints?

lib/db/RoutingDAO.js
```diff
@@ -0,0 +1,280 @@
...
+   * @param {number} intersectionTo - intersection to route to
+   * @returns {IntersectionRoutingResult?} best route between `intersectionFrom` and
+   * `intersectionTo`, excluding `intersectionFrom` but including `intersectionTo`,
+   * or `null` if no such route exists
+   */
+  static async routeIntersections(intersectionFrom, intersectionTo) {
+    if (intersectionFrom === intersectionTo) {
+      return { cost: 0, route: [] };
+    }
+    const sqlVertices = `
```
candu, line 44: can we extract this part out into a separate method here that returns the bounding box as an object `{ xmin, ymin, xmax, ymax }`, or `null` if one or both vertices are missing?

PR #527, opened 2020-07-14
Integrate corridor interpolation with frontend

### LINKED ISSUE #510: Multi-location: corridor interpolation in Edit mode

candu:
# Issue Addressed
This PR closes #510 .

# Description
We add new selection states `locationsSelection` / `locationsEditSelection` alongside `locations` / `locationsEdit`.  These selection states store the user's selected locations, as well as a `selectionType` that identifies whether this is a corridor or non-corridor selection.

In `MULTI_EDIT` mode, when the user updates the `selectionType` or the list of selected locations, we call a new action `syncLocationsEdit` that updates `locationsEdit`.  In non-corridor mode, it just copies the selected locations to `locationsEdit`; in corridor mode, it additionally makes a request to our routing API, then copies the resulting corridor to `locationsEdit`.

We also provide a new action `initLocations`, which initializes `locations` and `locationsSelection` from frontend route params.

Note that these changes do _not_ affect REST API endpoints for fetching collision / study data: the intent is that we can pass `CompositeId.encode(locations)` to these endpoints, so that from their point of view a corridor is just another set of centreline features to fetch data over.

# Tests
Tested the View Map and multi-location selection flows in frontend, including toggling the corridor checkbox on and off.

There is still a known issue where `routing.centreline_vertices` and `routing.centreline_edges` contain _all_ centreline features, including ones of types not captured in `RoadIntersectionType` / `RoadSegmentType`.  We'll need to think about what the right approach there is.

### DIFF
--- a/web/components/FcPaneMap.vue
+++ b/web/components/FcPaneMap.vue
@@ -90,7 +90,7 @@
 <script>
 import mapboxgl from 'mapbox-gl/dist/mapbox-gl';
 import Vue from 'vue';
-import { mapMutations, mapState } from 'vuex';
+import { mapGetters, mapMutations, mapState } from 'vuex';
 
 import { CentrelineType, LocationMode, MapZoom } from '@/lib/Constants';
 import { debounce } from '@/lib/FunctionUtils';
@@ -221,9 +221,6 @@ export default {
         this.setLegendOptions(legendOptions);
       },
     },
-    locationsForMode() {
-      return this.locationMode === LocationMode.MULTI_EDIT ? this.locationsEdit : this.locations;
-    },
     locationsGeoJson() {
       const features = this.locationsForMode.map(
         ({ geom: geometry, ...properties }) => ({ type: 'Feature', geometry, properties }),
@@ -235,6 +232,7 @@ export default {
     },
     locationsMarkersGeoJson() {
       const waypoints = this.locationsSelectionForMode.locations;
+      const n = waypoints.length;
 
       let waypointIndex = 0;
       let waypoint = waypoints[waypointIndex];
@@ -257,15 +255,19 @@ export default {
          * We could forbid selecting the same location multiple times, but that would introduce
          * a lot of validation complexity in both frontend and backend.  It would also make it
          * impossible to select a corridor that loops back on itself.
+         *
+         * `waypointIndices` represents the subsequence of `waypoints`, starting at the current
+         * `waypointIndex`, that matches the current location.
          */
         const waypointIndices = [];
-        while (waypointIndex < waypoints.length
+        while (waypointIndex < n
           && propertiesRest.centrelineType === waypoint.centrelineType
           && propertiesRest.centrelineId === waypoint.centrelineId) {
           waypointIndices.push(waypointIndex);
           waypointIndex += 1;
           waypoint = waypoints[waypointIndex];
         }
+        const k = waypointIndices.length;
 
         /*
          * `locationIndex === null` here indicates that the location is not a waypoint, and
@@ -273,7 +275,6 @@ export default {
          */
         let locationIndex = -1;
         let selected = false;
-        const k = waypointIndices.length;
         if (k === 0) {
           if (propertiesRest.centrelineType === CentrelineType.SEGMENT) {
             // We only show corridor markers at intersections.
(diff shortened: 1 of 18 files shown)


web/components/FcPaneMapPopup.vue
```diff
@@ -295,7 +300,8 @@ export default {
         if (this.featureLocationsEditIndex !== -1) {
           return false;
         }
-        return this.locationEditIndex === -1 && this.locationsEdit.length >= MAX_LOCATIONS;
+        return this.locationsEditIndex === -1
+          && this.locationsEditSelection.locations.length >= MAX_LOCATIONS;
```
candu, line 304: can we move this condition (the `length >= MAX_LOCATIONS` one) to a getter, e.g. `locationsEditFull`?  we can then use it here and in `FcPaneMapPopup`.

web/components/inputs/FcSelectorMultiLocation.vue
```diff
@@ -6,15 +6,15 @@
...
-            v-model="locationsEdit[i]"
+            v-model="locationsEditSelection.locations[i]"
             :location-index="i"
-            :selected="i === locationEditIndex"
+            :selected="i === locationsEditIndex"
             @focus="setLocationEditIndex(i)"
             @location-remove="actionRemove" />
           <FcInputLocationSearch
-            v-if="locationsEdit.length < MAX_LOCATIONS"
+            v-if="locationsEditSelection.locations.length < MAX_LOCATIONS"
```
candu, line 17: see previous comment - please use new getter here as well

web/components/inputs/FcSelectorMultiLocation.vue
```diff
@@ -153,37 +175,56 @@ export default {
     },
     messagesMaxLocations() {
       if (this.locationMode !== LocationMode.MULTI_EDIT
-        || this.locationsEdit.length < MAX_LOCATIONS) {
+        || this.locationsEditSelection.locations.length < MAX_LOCATIONS) {
```
candu, line 178: and here

web/components/FcPaneMap.vue
```diff
@@ -234,30 +234,87 @@ export default {
...
-          properties.selected = i === this.locationEditIndex;
+
+        /*
+         * It is possible for the current location to match multiple consecutive waypoints.
+         * We could forbid selecting the same location multiple times, but that would introduce
+         * a lot of validation complexity in both frontend and backend.  It would also make it
+         * impossible to select a corridor that loops back on itself.
+         */
+        const waypointIndices = [];
+        while (waypointIndex < waypoints.length
```
candu, line 262: can we set `const n = waypoints.length` above and use that here?  (this helps preserve the common `k` of `n` naming convention from various branches of mathematics, which would probably help make it clearer that `waypointIndices` identifies a subsequence of `waypoints`.  also we should say that explicitly in the comment here.)

candu: (I'd also love to get this under test, as it's an outlier among computed properties in terms of its complexity...but that will likely have to wait until we have spare cycles to build up some UI unit tests.)

web/components/FcPaneMap.vue
```diff
@@ -234,30 +234,87 @@ export default {
...
+        const feature = { type: 'Feature', geometry, properties };
+        features.push(feature);
       });
+
       return {
         type: 'FeatureCollection',
         features,
       };
     },
+    locationsSelectionForMode() {
```
candu, line 313: can we make this and `locationsForMode` getters as well?  after all, they depend exclusively on `vuex` state, and are likely to be used elsewhere once we start to build out View Data.

PR #550, opened 2020-07-30
Integrate analytics using the Oracle Infinity Data Collection API

### LINKED ISSUE #549: Investigate analytics integration using Data Collection API
**Description**
Previously we experimented with the City's installation of Oracle Infinity, and managed to get live page loads working properly.  That said, given that MOVE is a single-page application, we need a bit more fine-grained information than simply page loads.

This task investigates the [Data Collection API](https://docs.oracle.com/en/cloud/saas/marketing/infinity-user/Help/data_collection/dcapi.htm) as a promising way to integrate analytics more deeply within MOVE.

**Acceptance Criteria**
- [x] craft a `curl` command line that successfully sends an event via the Data Collection API, and verify that we can see that event in the dashboard;
- [x] based on information that needs to be passed there, decide whether DC API calls can be made from frontend or should be queued up in backend;
- [x] create a new module, `AnalyticsClient`, that sends along events as it receives them;
- [x] call `AnalyticsClient` from the following places:
  - [x] in the `created` hook to `App.vue`;
  - [x] in an `afterEach` hook to our Vue router;
  - [x] from a handful of action buttons (e.g. mode switches) that don't result in the previous changes;
- [ ] investigate event batching / debouncing approaches.

candu:
# Issue Addressed
This PR closes #549 .

# Description
We investigate the use of the [Oracle Infinity Data Collection API](https://docs.oracle.com/en/cloud/saas/marketing/infinity-user/Help/data_collection/dcapi.htm) to dynamically log events from MOVE.  This PR logs several types of events:

- all route transitions, including the initial page load;
- all button clicks;
- all clicks on elements of the left navigation bar;
- all sign in / sign out events;
- all search queries sent to the location suggestion API.

This was more complicated than originally supposed - the main wrinkles here were the [number of possible parameters](https://docs.oracle.com/en/cloud/saas/marketing/infinity-user/Help/parameters/reference.htm) to this API, and the difficulty of finding sets of parameters that result in events actually showing up in the dashboard.  Getting to a working set of events required us to carefully read through this documentation to find relevant parameters, observe `dcs.gif` requests sent by other City properties that use the `<script>` tag (e.g. the main toronto.ca website), craft `curl` requests to generate analytics events, and finally verify whether those events showed up in the dashboard.

(It didn't help that some of the documentation is flat out wrong or outdated.  When does that ever happen?  😬 )

Once that was done, we took those working `curl` requests and packaged them up into their `window.fetch()` equivalents in a new module `AnalyticsClient`.  Here there was another challenge: this client needs to both be accessible to components in our Vue app, and to have a reference to that Vue app internally to extract application state for logging - a circular dependency!  The solution was to export a singleton instance of `AnalyticsClient` from `@/web/analytics/analyticsClient`, extend `Vue.prototype` _before_ instantiating the Vue app to inject this instance into all components, and then finally call a `setAppContext` method to provide the app context to this client.

All in all: more work than expected, and there may still be rough edges around logging analytics events, but the path to making further changes is much better understood now.

# Tests
Tested individual events via `curl` / checking the dashboard.  Tested `AnalyticsClient` integration by checking in devtools Network tab that analytics requests are sent, and checking in the Oracle Infinity dashboard that those events can be viewed there.

### DIFF
--- a/web/main.js
+++ b/web/main.js
@@ -69,6 +69,12 @@ const vuetify = new Vuetify({
   },
 });
 
+/*
+ * Inject the singleton analytics client into all Vue components as `this.$analytics`.
+ * See https://vuejs.org/v2/guide/plugins.html#Writing-a-Plugin for why this works.
+ *
+ * Note that, at this point, `analyticsClient.appContext === null`.
+ */
 Object.defineProperty(Vue.prototype, '$analytics', {
   get() { return analyticsClient; },
 });
@@ -79,4 +85,17 @@ const appContext = new Vue({
   store,
   vuetify,
 }).$mount('#app');
+
+/*
+ * Once the application context has been created above, set that context in our
+ * singleton analytics client.
+ *
+ * Note that, due to the tick-based nature of Vue rendering, this will be called
+ * before any components actually render, and definitely before `router.afterEach()`
+ * is reached.  This allows us to ensure that `analyticsClient.appContext !== null`
+ * before any analytics events are sent.
+ *
+ * This two-stage init / set approach is necessary to avoid a circular dependency
+ * between the application context and the analytics client.
+ */
 analyticsClient.setAppContext(appContext);
--- a/web/main.js
+++ b/web/main.js
@@ -4,11 +4,12 @@ import Vuelidate from 'vuelidate';
 import Vuetify from 'vuetify/lib/framework';
 import en from 'vuetify/es5/locale/en';
 
-import App from '@/web/App.vue';
-import router from '@/web/router';
-import store from '@/web/store';
 import { formatDuration } from '@/lib/StringFormatters';
 import TimeFormatters from '@/lib/time/TimeFormatters';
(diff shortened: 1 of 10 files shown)


web/router.js
```diff
@@ -242,6 +243,10 @@ router.beforeEach(async (to, from, next) => {
       }
     } else {
       next(false);
+
+      const event = analyticsClient.signInEvent();
```
candu, line 247: with several of these events, we've made deliberate judgments about what sequence they should occur in.  we should either document those judgments at these callsites, or in `AnalyticsClient` itself (as some normative advice on how to use this).

web/main.js
```diff
@@ -68,9 +69,14 @@ const vuetify = new Vuetify({
   },
 });
 
-new Vue({
+Object.defineProperty(Vue.prototype, '$analytics', {
```
candu, line 72: the exact sequence of calls here is important.  can we document this too?

Issue #554, opened 2020-07-30
Multi-location: stub out scheduler service


candu:
**Description**
As described in [recent dev planning and estimation](https://www.notion.so/bditto/Dev-planning-and-estimation-for-View-Data-View-Reports-b6e822d200e44900b371f99a5cbadf00), background execution of report generation jobs is likely to require a new service, which we're calling `scheduler` in the docs.

This task creates a stub service for that, building out all the pieces needed to have this working (but not actually implementing any of them).  It also ensures that documentation and component diagrams are up to date after this and previous changes to system architecture.

**Acceptance Criteria**
- [x] create new `scheduler` directory with `scheduler.js` and `README` files;
- [x] copy over hapi.js setup to `scheduler.js`;
- [x] write quick blurb about `scheduler` in `README`, and update other documentation as necessary;
- [x] add new `JobController` with stub endpoints for creating jobs, checking job status, and cancelling jobs;
- [x] define port regions for `web`, `reporter`, and `scheduler` to avoid port collisions in multi-`reporter` setup, and update configs accordingly;
- [x] add new `/scheduler` reverse-proxy mapping to `nginx` configs;
- [x] update MOVE component diagram to reflect parallel `reporter` instances and new `scheduler` service.

candu:
Updated component diagram at https://www.notion.so/bditto/MOVE-Physical-Component-Diagram-2a35b14ae84c4f15b8758dad19dfb81a - marking closed.

PR #568, opened 2020-08-05
Stub server for MOVE Scheduler

candu:
# Issue Addressed
This PR makes significant progress on #554 .

# Description
We introduce `scheduler`, a new backend service for managing background jobs.  To provide more flexibility for future scaling, we assign each major component of MOVE its own `8x00-8x99` port range, with ports `9x00-9x99` used as debug / info ports in local development.

# Tests
Ran all system components in local development, tested that they can be accessed both via `webpack-dev-server` proxies and via direct port access.  Need to deploy this to test `forever` / `nginx` configuration.


scripts/dev/flashcrow.conf
```diff
@@ -1,6 +1,6 @@
 location /api {
```
candu, line 1: why do we even have this file?  we've never used `nginx` in local development...

PR #574, opened 2020-08-13
Report generation background job implementation

### LINKED ISSUE #558: Multi-location: implement report generation job
**Description**
In previous issues, we built the new `scheduler` service, and put scaffolding in place to support the intended report generation flow.

This task finishes that job by actually generating reports via HTTP calls to `reporter`.

**Acceptance Criteria**
- [x] decide on a strategy for storing reports:
  - [x] what should the path to a report be?
  - [x] what should the path to a ZIP archive be?
  - [x] what do we do in development?  in AWS without S3?  in AWS once we get S3?
- [x] create key-value storage layer for files;
- [x] change `BackgroundJobGenerateReports` as follows:
  - [x] for each `(type, id, format)`, generate a stub report and save that to disk;
  - [x] when generating a report, check first whether it has already been created;
  - [x] at the end, ZIP those responses into an archive;
  - [x] when generating a ZIP, check first whether it has already been created;
- [x] when generating individual reports, fetch the report data via HTTP call to `reporter`.

**Additional Notes**
While we could just call `reporter` libraries directly from within `scheduler`, this setup allows us to scale `reporter` independently of `scheduler`, and it gives us the benefit of load-balancing across `reporter` instances.

candu:
# Issue Addressed
This PR closes #558 .

# Description
We implement `JobRunnerGenerateReports` to generate reports via HTTP requests to `reporter`, then ZIP those reports together using `archiver`.

To support this implementation, we introduce `@/lib/io/storage`, a set of modules that offers a key-value store API for file storage.  This API uses `StorageStrategyBase` to allow for both filesystem-backed and S3-backed storage implementations, and also permits both streaming and `async` use by callers.  `StreamUtils` allows `async` callers to wait on streams as desired - this can be used with `await`, `Promise.all()`, etc. to allow for various concurrency flows.

We also update `@/lib/api`: `BackendClient` now uses `axios`, and has been renamed to `AxiosBackendClient` to reflect this.  This allows us to use the same client implementation in both frontend and backend.  Unit and REST API tests have been updated accordingly.  `httpsAgentLocal` provides an instance of `https.Agent` that accepts our local development certificates, for use in testing; we use `MoveConfig` to retrieve the `reporter` root URL and to determine whether `httpsAgentLocal` should be loaded.

# Tests
Lots of unit testing throughout the `@/lib/io/storage` library.  Updated REST API tests and ran using `npm run test:test-api` to ensure these still work.  Pre-commit hook tests.

Ran `scheduler` and tested various API calls using `curl`.  Checked that database `pg-boss` and job metadata records are generated correctly, that individual reports are saved to disk, that the ZIP file is saved to disk, that reports are actually PDFs, that the ZIP file is an actual ZIP file containing the actual reports, and that the filesystem directory hierarchy looks good.

Also checked that our changes to `BackendClient` didn't result in regressions by running `frontend`, `backend`, and `reports` and checking various flows: map navigation, login / logout, View Data, View Reports, download report.


lib/config/MoveConfig.js
```diff
@@ -119,13 +130,18 @@ const config = {
       issuerUrl: getAdfsIssuerUrl(DOMAIN_NAME),
     },
     port: 8100,
+    reporter: 'http://localhost/reporter',
```
candu, line 133: document that this uses the `upstream` block in `nginx` configs

lib/config/MoveConfig.js
```diff
@@ -119,13 +130,18 @@ const config = {
...
     session: {
       /*
        * In production, we need the cookie to be tied to our ELB domain
        * name, not localhost.
        */
       domain: DOMAIN_NAME,
     },
+    storage: {
+      type: StorageType.FILESYSTEM,
+      args: ['/data/move-storage'],
```
candu, line 143: document that this uses the EFS mount?

lib/io/StreamUtils.js
```diff
@@ -0,0 +1,46 @@
+import stream from 'stream';
```
candu, line 1: these seem OK, but can we add some quick tests?  (e.g. valid empty stream, valid non-empty stream, stream with error)

lib/io/storage/StorageHash.js
```diff
@@ -0,0 +1,13 @@
+import crypto from 'crypto';
+
+class StorageHash {
```
candu, line 3: quick documentation why this is here, please

lib/io/storage/StoragePath.js
```diff
@@ -0,0 +1,28 @@
+import crypto from 'crypto';
+
+class StoragePath {
```
candu, line 3: same here: just quickly document what this does

lib/test/api/InjectBackendClient.js
```diff
@@ -1,36 +1,43 @@
-import BackendClient from '@/lib/api/BackendClient';
+import QueryString from '@/lib/api/QueryString';
```
candu, line 1: we should probably test `InjectBackendClient` as well, similar to our unit tests for `AxiosBackendClient`

PR #584, opened 2020-08-25
Integrate background report generation with frontend

### LINKED ISSUE #560: Multi-location: connect report generation frontend to scheduler
**Description**
As with other features, this task involves adding new `JobController` endpoints to `WebApi`, and then using those client calls from Aggregate View.

**Acceptance Criteria**
- [x] `JobController` endpoints are added to `WebApi`;
- [x] call `WebApi` job creation method from Aggregate View study reports when download is requested;
- [x] when snackbar appears, change it to persist with no timeout, and poll `WebApi` status-checking method to update text;
- [x] once job is complete, change snackbar to "done" status, and allow user to download the result.

**Additional Notes**
This may actually be a good task to do after #557 , as it will allow us to more easily run end-to-end tests of background jobs as we work on #558 .

candu:
# Issue Addressed
This PR closes #560 .

# Description
We integrate backgroudn report generation via `scheduler` / `reporter` in the aggregate view within View Data, adding new buttons / modes to match designs and connecting those to new methods in `WebApi` that call relevant API endpoints.

We also add `FcToastJob`, which polls `GET /job/{id}` in the `scheduler` service to check on the status of the given job.  This toast offers job state-dependent actions: undo (i.e. cancel) when the job is in progress, download once it has completed.  Downloads are powered by the new `StorageController`, which returns the result of `storageStrategy.getStream()` in the response and uses `mime-types` to infer MIME type from the storage `key`.

Note that this does _not_ yet include any of the selection / filter modal parts of this feature from the designs; those are coming soon!

# Tests
Tested bulk reporting modes in the frontend for both collision and study reports.  Added unit tests for directory traversal mitigation in `StorageStrategyFilesystem`, as well as for `null` parameters in `QueryString`.

### DIFF
--- a/lib/api/WebApi.js
+++ b/lib/api/WebApi.js
@@ -194,6 +194,14 @@ async function getReportWeb(type, id, reportParameters) {
   };
 }
 
+async function getStorage(namespace, key) {
+  const options = {
+    method: 'GET',
+    responseType: 'blob',
+  };
+  return apiClient.fetch(`/storage/${namespace}/${key}`, options);
+}
+
 async function getStudiesByCentreline(features, studyType, filters, pagination) {
   const s1 = CompositeId.encode(features);
   const data = { s1, ...filters, ...pagination };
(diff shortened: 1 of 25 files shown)


lib/api/WebApi.js
```diff
@@ -181,6 +194,14 @@ async function getReportWeb(type, id, reportParameters) {
   };
 }
 
+async function getStorage(namespace, key) {
+  const options = {
+    method: 'GET',
+    responseType: 'blob',
```
candu, line 200: it would be really nice to make these client-side API wrappers streaming, but we can deal with that later.

lib/controller/StorageController.js
```diff
@@ -0,0 +1,50 @@
...
+    validate: {
+      params: {
+        namespace: Joi.string().required(),
+        key: Joi.string().required(),
+      },
+    },
+  },
+  handler: async (request, h) => {
+    const { namespace, key } = request.params;
+    const fileExists = await storageStrategy.has(namespace, key);
```
candu, line 34: use `Boom.boomify` to handle errors on `storageStrategy.has()` here

Issue #562, opened 2020-07-31
Multi-location: flesh out download manager


candu:
**Description**
In #561 , we created a new route and associated interface for the "download manager", or notifications area.

This task builds on that, replacing the stub interface with actual download job information.

**Acceptance Criteria**
- [x] add a method to `JobMetadataDAO` that fetches download jobs for the current user;
- [x] add an endpoint to `JobController` that fetches metadata and job records for the current user's download jobs;
- [x] using `WebApi`, use this endpoint in the download manager to fetch the list of jobs;
- [x] in the download manager frontend:
  - [ ] sort jobs: first by status (in progress / queued / completed), then by most recent;
  - [x] for jobs in progress, show progress bar and time estimate;
  - [x] for queued jobs, show a "queued" state;
  - [x] for completed jobs that succeeded, show a download link;
  - [ ] for completed jobs that failed, show a retry link.

**Additional Notes**
Unlike #561 , this _will_ have to wait until the rest of the download job infrastructure is in place.

candu:
Going to accept this as done - we should probably get more eyes on it before making any further decisions on how /whether to organize jobs.

candu:
PR #585 addressed this - marking as closed.

PR #585, opened 2020-08-26
Download manager page

### LINKED ISSUE #561: Multi-location: stub out download manager
**Description**
In previous tasks, we scaled out `reporter`, built the `scheduler` service, created a report generation background job, and tied that job in with the frontend.

However, we haven't yet provided users with a way to come back to old download jobs!  Figma designs call for a new "notifications" area, reachable via the left navbar, to contain a record of report generation jobs that the user has submitted.

This task starts the implementation of that new area, which we'll call (for now) the "download manager".

**Acceptance Criteria**
- [x] create new route for download manager in `router.js`;
- [x] add new left nav icon and link it to this new route;
- [x] sketch out rough interface for download manager;
- [x] stub out that interface in the download manager.

**Additional Notes**
This task doesn't actually depend on anything else, and could be implemented immediately - I'd recommend shortly after #553 and #554 are deployed and tested.

candu:
# Issue Addressed
This PR closes #561 and #562 .

# Description
We introduce the MOVE Download Manager (`FcDownloadsManage.vue`), a new top-level page in MOVE where the user can check download status and download previously requested bulk report archives.

To power this interaction, we refactor the job status-polling logic out into a new module `JobPoller`, which is now used from both `FcToastJob` and `FcCardJob`.

We also implement rudimentary read / unread states with the `dismissed` flag in `JobMetadata`.

# Tests
Updated unit and database tests as appropriate.  Tested flows manually in frontend.

### DIFF
--- a/lib/jobs/JobType.js
+++ b/lib/jobs/JobType.js
@@ -34,6 +34,17 @@ JobType.init({
       dataNormalized.s1 = s1;
       return dataNormalized;
     },
+    async validateMetadata(metadata) {
+      const metadataSchema = Joi.object().keys({
+        ...CentrelineSelection,
+        reportExportMode: Joi.enum().ofType(ReportExportMode),
+      });
+
+      const { s1 } = metadata;
+      const metadataNormalized = await metadataSchema.validateAsync(metadata);
+      metadataNormalized.s1 = s1;
+      return metadataNormalized;
+    },
   },
 });
 JobType.NAME_PREFIX = 'MOVE:';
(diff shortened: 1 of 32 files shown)


lib/jobs/JobType.js
```diff
@@ -1,30 +1,50 @@
...
+            type: Joi.enum().ofType(ReportType).required(),
+            id: Joi.string().required(),
+            format: Joi.enum().ofType(ReportFormat).required(),
+          }).unknown(),
+        ),
+      });
+
+      const { s1 } = data;
+      const dataNormalized = await dataSchema.validateAsync(data);
+      dataNormalized.s1 = s1;
```
candu, line 34: there's a few instances in this PR where we have to specially handle `CompositeId` values, all because our `Joi.compositeId()` extension changes the value from a string to an array.

just submitted this as #586 - we can address that separately at some point.  (it would also allow us to simplify the logic here: we could just define `dataSchema`, `metadataSchema` as properties and use those directly.)

PR #587, opened 2020-08-27
Bulk downloads: notification dot, authentication, prioritization

### LINKED ISSUE #563: Multi-location: notify user of newly available downloads
**Description**
In #561 , we added a "download manager" area of MOVE, and gave it its own icon on the left navbar.

This task brings newly available downloads to the user's attention through a dot on the left navbar icon, as well as by prioritizing new downloads in the download manager.

**Acceptance Criteria**
- [x] in job metadata, add a new column `BOOLEAN dismissed`;
- [x] when the user downloads the ZIP file from a completed job, mark this flag `true`;
- [x] in the download manager itself, bring completed and undismissed jobs to the top;
- [x] add a method to `JobMetadataDAO` that efficiently determines if the user has any completed and undismissed jobs;
- [x] add a REST API endpoint that calls this DAO method;
- [x] add DAO and REST API tests as needed;
- [x] call this endpoint during initial page load via `WebApi`;
- [x] if this endpoint returns `true`, show a dot on the icon.

**Additional Notes**
We may eventually wish to show a similar dot for Track Requests, so it will likely be useful to build this functionality directly into navbar icon components.

candu:
# Issue Addressed
This PR closes #563 .

# Description
We add a new `GET /jobs/existsNew` endpoint to `scheduler`, and call this endpoint on page load to determine whether to show a "new downloads" notification dot in the left navbar.

We also add auth / scope check functionality to `FcButton`, then use this functionality to force login when an unauthenticated user attempts to enter bulk report mode.  (Even if they could enter bulk report mode, the relevant backend endpoints would return HTTP 403, but it's still good practice.)

Finally, download jobs in the download manager are prioritized to bring in-progress and newly completed jobs to the top.  This helps the user quickly find the most salient details.  (For now, we don't do any pagination on the list of download jobs - we can always add that later!)

# Tests
Added tests for DAO method that powers `GET /jobs/existsNew` in `JobMetadataDAO.spec.js`.  Tested flows manually in frontend.

### DIFF
--- a/web/components/inputs/FcButton.vue
+++ b/web/components/inputs/FcButton.vue
@@ -12,6 +12,11 @@
 </template>
 
 <script>
+import { mapMutations, mapState } from 'vuex';
+
+import { hasAuthScope } from '@/lib/auth/ScopeMatcher';
+import { saveLoginState } from '@/web/store/LoginState';
+
 const BUTTON_ATTRS = {
   primary: {
     color: 'primary',
@@ -41,12 +46,17 @@ const BUTTON_ATTRS = {
 export default {
   name: 'FcButton',
   props: {
+    scope: {
+      type: Array,
+      default: null,
+    },
     type: String,
   },
   computed: {
     typeAttrs() {
       return BUTTON_ATTRS[this.type];
     },
+    ...mapState(['auth']),
   },
   methods: {
     actionClick() {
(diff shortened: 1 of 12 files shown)


web/components/inputs/FcButton.vue
```diff
@@ -59,7 +69,31 @@ export default {
...
+      }
+      if (loggedIn) {
+        if (hasAuthScope(user, scope)) {
+          return true;
+        }
+        this.setDialog({
+          dialog: 'ConfirmUnauthorized',
+          dialogData: { scope },
+        });
+        return false;
```
candu, line 87: I wish this would cancel other event handlers, but sadly that appears to not be the case - so for buttons that visibly change the interface (such as our export / cancel toggles in Aggregate View), the user will see a brief flash of the new state before being redirected to login.

Issue #599, opened 2020-09-02
Multi-location: stub out wizard stepper and pages for Request Study


candu:
**Description**
#598 started scaffolding out the implementation of multi-location Request Study.  This task continues that by scaffolding out the steps and corresponding pages of the multi-location Request Study wizard.

**Acceptance Criteria**
- [x] show wizard stepper at top
- [x] show bottom bar actions as appropriate for each step, and wire them up to change steps;
- [x] scaffold pages:
  - [x] page 1 (intersections)
  - [x] page 2 (midblocks)
  - [x] page 3 (summary details)
  - [x] page 4 (review and confirm)
  - [x] end state

**Additional Notes**
For now, we don't need to make any attempt to show actual data; we can cover that in upcoming tasks, and just use copy from the designs for now.  We'll also get to actually saving this request later on.

PR #610, opened 2020-09-03
Stub modes for multi-location Request Study

### LINKED ISSUE #598: Multi-location: stub out states for new Request Study page
**Description**
For a while, we've disabled Request Study while we work on other multi-location features.  This task starts the process of building a new multi-location flow for Request Study, and of re-integrating that with the rest of MOVE.

**Acceptance Criteria**
- [x] change URL for Request Study page to include location selection details, and make it call `initLocations` in `loadAsyncForRoute`;
- [x] reconnect "Request New..." button in View Data to Request Study page;
- [x] handle modes in Request Study: (<2h)
  - [x] if in `LocationMode.SINGLE` or `detailView`, show single-location request flow (as it is now);
  - [x] otherwise, show bulk request flow (just a TODO stub for now).

candu:
# Issue Addressed
This PR closes #598 .

# Description
We refactor `FcDrawerRequestStudy` to handle two sets of modes: create vs. edit, and bulk vs. non-bulk.  To facilitate this, we add location selection parameters to the route for Request Study, and use those with `initLocations()` to determine whether to load the bulk or non-bulk interface.

For now, the bulk interface is just a `<h2>` with a TODO message - we'll be filling that out as we work on #599 and beyond.

# Tests
Tested quickly in frontend - testing will get more stringent as we proceed through these tasks.

### DIFF
--- a/web/components/requests/FcDetailsStudyRequest.vue
+++ b/web/components/requests/FcDetailsStudyRequest.vue
@@ -445,38 +445,6 @@ export default {
         this.internalValue.dueDate = this.dueDate;
       }
     },
-    location() {
-      this.updateStudyRequestLocation();
-    },
-  },
-  created() {
-    this.updateStudyRequestLocation();
-  },
-  methods: {
-    updateStudyRequestLocation() {
-      if (this.location === null) {
-        this.studyRequest.centrelineId = null;
-        this.studyRequest.centrelineType = null;
-        this.studyRequest.geom = null;
-      } else {
-        const {
-          centrelineId,
-          centrelineType,
-          lng,
-          lat,
-        } = this.location;
-        const geom = {
-          type: 'Point',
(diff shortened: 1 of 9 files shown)


web/components/requests/FcDetailsStudyRequest.vue
```diff
@@ -408,6 +445,38 @@ export default {
         this.internalValue.dueDate = this.dueDate;
       }
     },
+    location() {
+      this.updateStudyRequestLocation();
```
candu, line 449: can we get rid of this?  i don't think we want to allow location editing in Request Study - which also means that we'll need to lock / hide certain features on `FcPaneMap` (but we can do that part in #599 ).

candu (commented):
remove location updates, otherwise good

Issue #292, opened 2020-01-09
Set up QA environment for MOVE (web)


candu:
**Description**
So far, we have the `flashcrow-dev0` AWS stack as provisioned by CSG.  This stack is used for all deployments, including incremental mid-sprint changes.

Having an additional QA environment in `ca-central-1` would give us a more stable environment to put in front of users, especially as we study longer-term usage patterns and viability of CRASH / FLOW replacement.

**Acceptance Criteria**
- [x] CSG to provision new QA environment on AWS in `ca-central-1`, with similar configuration to `flashcrow-dev0` stack;
- [x] CSG to provide SSH keys, domains, and other relevant details to MOVE;
- [x] MOVE to prepare application, scripts, and configuration for deployment to QA;
- [ ] MOVE to deploy and verify setup, and coordinate with CSG for changes as needed.

candu:
@mkewins @aharpalaniTO here's the QA environment GitHub issue, in case it helps anchor conversations around this with CSG / IT.

candu:
Marking this closed - I'll create a separate issue for the ETL / Airflow side of things here.

PR #614, opened 2020-09-10
Update CI / CD scripts as per recent QA environment discussions

candu:
# Issue Addressed
This PR makes further progress on #292 .

# Description
We switch away from using EFS to store application secrets, in favour of heavier use of SSM.

# Tests
Going to test this by running the deployment process (and possibly hotfixing master if it doesn't work).


candu (commented):
LGTM

Issue #608, opened 2020-09-02
Multi-location: new View Request layout


candu:
**Description**
Since we now have our new View Bulk Request page from #607 , we also need to update the View Request page to match that look-and-feel (and our designs).  This task does just that.

**Acceptance Criteria**
- [x] implement new side-by-side layout for View Request;
- [x] back button:
  - [x] if request is part of a bulk request, goes to that bulk request;
  - [x] otherwise, goes back to either View Data or Track Requests as appropriate;
- [x] improved status actions (see #448 ).

PR #623, opened 2020-09-21
Edit actions for bulk and non-bulk study requests

### LINKED ISSUE #607: Multi-location: View Bulk Request actions
**Description**
In #606 , we scaffolded out the new View Bulk Request page.  This task builds on that by implementing the status / assigned to actions in that page.

**Acceptance Criteria**
- [x] add selection to table of child requests;
- [x] add status and assigned to dropdowns;
- [x] disable those dropdowns when nothing selected;
- [x] disable options in those dropdowns that cannot be applied to all selected requests;
- [x] when selecting from those dropdowns, apply corresponding action to all selected.

candu:
# Issue Addressed
This PR closes #607 and makes further progress on #608 .

# Description
We introduce `FcMenuStudyRequestStatus`, a component for managing status change actions in both bulk and non-bulk study requests.  We then use that component from within the bulk and non-bulk View Request flows.

For bulk requests, this is powered by the new `PUT /request/study/bulk/{id}` endpoint.

To simplify the frontend and reduce the number of interface states, we also change both bulk and non-bulk `PUT` endpoints to _not_ return `StudyRequestChange` objects as well.  As part of this, status change actions now result in a full data reload via `loadAsyncForRoute` - this does result in more data fetching, but is significantly simpler.

# Tests
Tested in frontend.  Updated `test:test-db` and `test:test-api` test suites as needed and ran these successfully.

### DIFF
--- a/lib/auth/StudyRequestPermissions.js
+++ b/lib/auth/StudyRequestPermissions.js
@@ -3,6 +3,16 @@ import Boom from '@hapi/boom';
 import { AuthScope, StudyRequestStatus } from '@/lib/Constants';
 import { hasAuthScope } from '@/lib/auth/ScopeMatcher';
 
+/**
+ * Checks the details of a study request to ensure that we're not trying to update
+ * non-updatable columns.  Works for both bulk and non-bulk requests.
+ *
+ * @param {Object} studyRequestNew - new study request object, usually as passed in the payload
+ * of a `PUT` request
+ * @param {Object} studyRequestOld - old study request object, usually as fetched using
+ * {@link StudyRequestDAO.byId}
+ * @returns {hapi.Boom?} `Boom` error if the update cannot proceed, or `null` if it can
+ */
 function canUpdateStudyRequestDetails(studyRequestNew, studyRequestOld) {
   if (studyRequestNew.id !== studyRequestOld.id) {
     return Boom.badRequest('cannot change ID for study request');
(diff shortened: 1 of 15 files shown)


lib/auth/StudyRequestPermissions.js
```diff
@@ -0,0 +1,101 @@
...
+  return null;
+}
+
+function canUpdateStudyRequest(studyRequestNew, studyRequestOld, user) {
+  const err = canUpdateStudyRequestDetails(studyRequestNew, studyRequestOld);
+  if (err !== null) {
+    return err;
+  }
+
+  if (studyRequestNew.status !== studyRequestOld.status
```
candu, line 35: check that `studyRequestBulkId` doesn't change either

web/components/requests/status/FcMenuStudyRequestsStatus.vue
```diff
@@ -0,0 +1,236 @@
...
+          :key="i"
+          class="fc-item-study-requests-status"
+          :disabled="item.disabled"
+          @click="actionMenu(item, null)">
+          <v-list-item-title>
+            <v-icon :color="item.value.color" left>mdi-circle-medium</v-icon>
+            <span>{{item.text}}</span>
+          </v-list-item-title>
+        </v-list-item>
+        <v-list-group
```
candu, line 31: right now, these groups (well, group, since there's only 1) stay open when the parent menu is closed.  we'd ideally like for them to close with the parent menu!

one way to do that is to keep a mapping from `item.value` to flags that determine whether each submenu is open - we can then add a watcher to set these flags to `false` whenever `showMenu` becomes `false`.

candu (commented):
couple of changes, otherwise good

Issue #628, opened 2020-09-22
Move ETL scripts to separate repository


candu:
**Description**
ETL-related scripts currently live alongside web application code in `bdit_flashcrow`.  In preparation for automated provisioning / deployment of Airflow, however, we should move these to a separate repository - this will help us deploy Airflow separately from the MOVE web application.

This task carries that migration out.

**Acceptance Criteria**
- [x] a new repo `bdit_move_etl` is created on GitHub;
- [x] generic language / linter / etc. configuration is added into `bdit_move_etl`;
- [x] ETL scripts and configuration are moved into `bdit_move_etl`;
- [x] our existing ETL setup is reconfigured to use a clone of `bdit_move_etl` for Airflow scripts and ETL configuration;
- [x] ETL scripts and configuration are removed from `bdit_flashcrow`;
- [x] `bdit_move_etl` has CodeDeploy configurations:
  - [x] `appspec.yml`, very similar to that for `bdit_flashcrow`;
  - [x] `deploy_scripts`, containing deployment lifecycle scripts for ETL / Airflow;
- [x] `bdit_move_etl` GitLab setup:
  - [x] copy repo from GitHub;
  - [x] set up dual-push;
  - [x] add GitLab CI file;
- [x] follow up with Cloud Services and @radumas re: changes and next steps.

candu:
Just followed up with Cloud Services - will file another issue to track progress on the ETL installation.

PR #631, opened 2020-09-23
Remove Airflow / ETL scripts and configs from MOVE web repo

candu:
# Issue Addressed
This PR makes progress on #628 .

# Description
We remove scripts and configs from `bdit_flashcrow` that relate to Airflow or ETL environment setup - these have been moved to `bdit_move_etl`.

With those gone, `bdit_flashcrow` no longer has any Python code to speak of, and so we can get rid of Python-related tooling as well!

# Tests
This will be tested post-merge via the GitLab CI / CD runner.


candu (commented):
LGTM

PR #648, opened 2020-10-05
Bulk actions: status management, assign to, notification of partial completion

### LINKED ISSUE #646: Beta prep: bulk request management actions
**Description**
There's a good deal of confusion around bulk request management actions - what should the individual row "Assigned To" dropdowns do?  why aren't bulk actions available in Track Requests?  why does the button disable itself and not let me change everything in the bulk request?

This task is part team discussion, part action item list for resolving this.

**Acceptance Criteria**
for discussion:
- [x] None in Assigned To: *should* unassigning be possible?  (tradeoffs: "undo-ability" vs. simplicity)
- [x] individual row "Assigned To" dropdowns?  (are these helpful or confusing?)
- [x] should we have bulk actions on Track Requests?
- [x] how to signal that bulk actions can't be performed (e.g. because they don't apply to all of the selected requests)?
  - [x] is this even a thing we want to do?  (for instance: should we just let users move a request directly from Requested to Completed if they want to?  and back again?)

for action:
- [x] keep "None" in "Assign To", but rename it to "Unassigned" for clarity;
- [x] "Assign To" should _always_ work in Requested, Changes Needed, and Assigned statuses:
  - [x] individual row dropdowns;
  - [x] status menu;
- [x] in bulk status dropdown, enable actions that can be performed on _any_ selected request (as opposed to _all_ such requests);
- [x] when an action that can be applied to _some_ but not _all_ selected requests is selected:
  - [x] change requests that can be changed;
  - [x] notify the user (e.g. via dialog) of any requests that could not be changed, and why.
- [x] add separate "Assign To" menu beside status menu in View Bulk Request;
- [x] add status menu and "Assign To" menu to Track Requests;
- [x] centralize permissions logic here where possible.

### LINKED ISSUE #647: Beta prep: User filter broken in Track Requests
**What Happened**
When visiting Track Requests with a user that does not have `STUDY_REQUESTS_ADMIN` permission, Track Requests just shows the loading spinner indefinitely - no data / table is ever loaded.

**What Should Happen**
The Track Requests page should be loaded as expected.  The user filter should remain active by default, and the user should see a list of only those requests they have submitted themselves.  All users should be able to enable / disable the User filter without issue.

**To Reproduce**
Steps to reproduce the bug:
1. Log in as a user without `STUDY_REQUESTS_ADMIN` permission (or use an `ADMIN` user and temporarily disable this permission for yourself via the Admin Console!)
2. Go to Track Requests.

At this point, Track Requests hangs indefinitely.

Similar behaviour is observed if you log in as _any_ user and select the "Requested By: User" filter in Track Requests - the table never updates, and the same stacktrace (see below) is logged to console.

**Additional Notes**
Stacktrace from AWS dev:

```
vue.runtime.esm.js:1888 TypeError: Cannot read property 'id' of undefined
    at P (RequestFilters.js:122)
    at R (RequestFilters.js:148)
    at FcRequestsTrack.vue:214
    at Array.map (<anonymous>)
    at o.items (FcRequestsTrack.vue:214)
    at nr.get (vue.runtime.esm.js:4479)
    at nr.evaluate (vue.runtime.esm.js:4584)
    at o.items (vue.runtime.esm.js:4836)
    at o (FcRequestsTrack.vue?2f84:1)
    at o.e._render (vue.runtime.esm.js:3548)
```

candu:
# Issue Addressed
This PR closes #646 (despite the branch name - oops!), and closes #647 .

# Description
We refactor common logic around study request actions out into `RequestActions`, then use that to change the semantics of the "set status" and "assign to" menus.  Previously these menus only showed actions that could be performed on _all_ selected study requests; now they show actions that can be performed on _any_ selected study request.  Since this can result in partial completion of a bulk action (e.g. if the user tries to "assign to" on a mix of Requested and Completed requests), we add a notification dialog to inform the user of which study requests could not be updated.

We then turn around and use these bulk action menus inline in the table, as well as in both Track Requests and View Bulk Request.  To prevent confusion / viewing stale states, we force a full refresh whenever requests are updated.

Finally, Track Requests gets a bugfix for #647 plus better selection!  Bulk request selector checkboxes now function as "select all" for the study requests that are part of that bulk request, with indeterminate states and everything.

# Tests
Tested Track Requests, View Bulk Request, and the individual row menus to make sure we could modify study requests.

### DIFF
--- a/lib/requests/RequestActions.js
+++ b/lib/requests/RequestActions.js
@@ -1,0 +1,100 @@
+import { AuthScope, StudyRequestStatus } from '@/lib/Constants';
+import { hasAuthScope } from '@/lib/auth/ScopeMatcher';
+
+class RequestActions {
+  /* eslint-disable no-param-reassign */
+  static actionAssignTo(studyRequest, assignedTo) {
+    studyRequest.assignedTo = assignedTo;
+    if (assignedTo === null) {
+      studyRequest.status = StudyRequestStatus.REQUESTED;
+    } else {
+      studyRequest.status = StudyRequestStatus.ASSIGNED;
+    }
+  }
+
+  static actionCancel(studyRequest) {
+    studyRequest.status = StudyRequestStatus.CANCELLED;
+    studyRequest.closed = true;
+  }
+
+  static actionMarkCompleted(studyRequest) {
+    studyRequest.status = StudyRequestStatus.COMPLETED;
+    studyRequest.closed = true;
+  }
+
+  static actionRejectData(studyRequest) {
+    studyRequest.status = StudyRequestStatus.REJECTED;
+    studyRequest.closed = true;
+  }
+
+  static actionReopen(studyRequest) {
+    studyRequest.status = StudyRequestStatus.REQUESTED;
+    studyRequest.closed = false;
+  }
+
+  static actionRequestChanges(studyRequest) {
+    studyRequest.status = StudyRequestStatus.CHANGES_NEEDED;
+  }
+  /* eslint-enable no-param-reassign */
+
+  static canAssignTo(user, studyRequest) {
+    if (!hasAuthScope(user, AuthScope.STUDY_REQUESTS_ADMIN)) {
+      return false;
+    }
+    const { status } = studyRequest;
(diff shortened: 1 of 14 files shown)


candu (commented):
LGTM

PR #658, opened 2020-10-08
a11y: keyboard navigability

### LINKED ISSUE #632: a11y: improve keyboard navigability
**Description**
During internal testing with the NVDA screen reader, several issues around keyboard navigation, tab ordering, and autofocus were discovered.  This task looks to address those.

**Acceptance Criteria**
- [x] when logged in, user menu should be keyboard navigable;
- [x] location selection:
  - [x] when searching for a location, location search results should be read out as they are focused;
  - [x] when a single location is selected and the user has not entered View Data mode or has collapsed the View Data drawer, the map popup should always show up (so that the "View Data" button can be tabbed to);
  - [x] after clicking "Add Location", focus should go to the "Search for or click..." input box;
- [x] status menu: trying to open the "Assign To" submenu closes the whole menu!
- [x] `FcDataTableRequests`: checkboxes (both "Select All" and individual row) not in tab order!

**Additional Notes**
There are specific problems with the Bulk Request Study wizard and Due Date field that will be covered in separate issues.

candu:
# Issue Addressed
This PR closes #632 .

# Description
We make several parts of MOVE much more easily navigable by keyboard: the user menu, map popups, bulk and non-bulk request study flows, the location search bar and multi-location selector.

We also address a small bug in the bulk request flow where it would show an empty list for selections that contain no intersections / midblocks in the corresponding wizard step.  Those steps are now disabled in the stepper at top, and are skipped by both forward and backward navigation.

# Tests
Tested changes in frontend.

### DIFF
--- a/web/App.vue
+++ b/web/App.vue
@@ -13,6 +13,7 @@
       v-bind="toastData" />
     <v-navigation-drawer
       app
+      class="fc-navigation-drawer"
       mini-variant
       permanent>
       <template v-slot:prepend>
@@ -117,6 +118,10 @@ export default {
   font-size: 0.875rem;
   font-weight: normal;
   line-height: 1.25rem;
+
+  & .fc-navigation-drawer {
+    overflow: visible;
+  }
 
   & .v-input--selection-controls__input + .v-label {
     color: var(--v-default-base);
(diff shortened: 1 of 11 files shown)


candu (commented):
LGTM

PR #659, opened 2020-10-09
a11y: misc cleanup

### LINKED ISSUE #637: a11y: miscellaneous issues
**Description**
Adding this as a catch-all issue for a11y problems not captured in previous issues.

**Acceptance Criteria**
- [x] contrast on disabled breadcrumbs in View Request (bulk and non-bulk) - needs to be higher;
- [x] get rid of `min-height` on `FcPaneMap`, just add this directly in View Request page (since this breaks View Reports layout).

candu:
# Issue Addressed
This issue closes #637 .

# Description
We fix a few last low-hanging a11y issues: marking disabled breadcrumbs in View Request flows with `aria-disabled`, initial focus on the location search bar, better popup handling, better focus management in location selection.

# Tests
Tested quickly in frontend.

### DIFF
--- a/web/components/FcDataTableRequests.vue
+++ b/web/components/FcDataTableRequests.vue
@@ -166,7 +166,7 @@
         <FcButtonAria
           :aria-label="item.ariaLabel"
           button-class="btn-show-request"
-          top
+          left
           type="secondary"
           @click="actionShowItem(item)">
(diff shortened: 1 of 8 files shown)


candu (commented):
LGTM

PR #665, opened 2020-10-13
Tweak map controls layout for better visibility on 1280x720

### LINKED ISSUE #455: Legend overlaps report menus on smaller screens
**What Happened**
@jwrcoleman @mkewins @shinechaudhuri have all previously noted an issue where the map legend overlaps report menus (e.g. selecting the study date) in View Reports when using MOVE on smaller screens.

**What Should Happen**
The legend and report controls should not conflict.

**To Reproduce**
Steps to reproduce the bug:
1. On the map, navigate to any location with available studies and click "View Data";
2. Click "View Reports" beside any of the study types;
3. Resize your browser window, reducing the height until the issue appears.

**Screenshots**
From @jwrcoleman , the bug in action:

image (6)

Note that the legend is _also_ obscuring several of the bottom-right map controls.

What it looks like on larger screens:

image

**Additional Notes**
As noted, we can mimic having a smaller screen by resizing the browser window.  The effective height of the window is given by `window.innerHeight`.  Some rough breakpoints here:

- `window.innerHeight === 875` is roughly the smallest height at which the layout still looks "normal" with no overlap;
- past that, the legend starts to overlap more and more of the map controls;
- at `window.innerHeight === 632`, the legend is starting to bleed down into the report drawer.

Similar to usual responsive breakpoints, which depend on window width, we can also [set breakpoints on height](https://stackoverflow.com/questions/43786717/vertical-responsive-height) - this is one potential way of fixing this.

candu:
# Issue Addressed
This PR closes #455 .

# Description
Moved the scale and map data attribution to bottom-left, which allows us to bump the rest of the controls down and de-clutters that bottom-right area.  Added `max-height` on the study date dropdown menu, so that it doesn't overlap the map even in vertical split-screen mode.  Finally, added a media query in `FcLayoutDrawerMap` that applies `max-height: 192px` and `overflow: auto` to the legend on screens with a height of 900px or less.

With these changes, we should at the very least minimize conflict between these controls even on 1280x720 (which is the smallest resolution we're seeing in our analytics so far).

# Tests
Tested using device toolbar in Chrome devtools to emulate a 1280x720 screen - navigated around the site a bunch with that.

### DIFF
--- a/web/components/FcDrawerViewStudyReports.vue
+++ b/web/components/FcDrawerViewStudyReports.vue
@@ -32,7 +32,7 @@
             </span>
             <v-menu
               v-else
-              max-height="320">
+              :max-height="320">
               <template v-slot:activator="{ on, attrs }">
                 <FcButton
                   v-bind="attrs"
@@ -65,7 +65,9 @@
             </v-chip>
           </div>
           <v-spacer></v-spacer>
-          <v-menu v-if="studies.length > 0">
+          <v-menu
+            v-if="studies.length > 0"
+            :max-height="320">
             <template v-slot:activator="{ on, attrs }">
               <FcButton
                 v-bind="attrs"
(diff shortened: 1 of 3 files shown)


candu (commented):
LGTM

PR #667, opened 2020-10-14
Direction-swapping bug with 3-way ("T") intersections

### LINKED ISSUE #666: Minor street approaches incorrect at T-intersection: Chesswood and Champagne
**What Happened**
From the MOVE Bug Report form:

> for Chesswood and Champagne intersection count the total volume and the major volume is the same.

Related to this, the table in the Intersection Summary Report shows that there are zero minor street approaches, which is at odds with the turning movements described in that same table.

**What Should Happen**
The total and major volumes should not be identical at this location, as there should be a non-zero number of minor street approaches.  (In general, this should be _very_ unlikely, unless for some reason the count was taken when the minor street was completely closed off.)

**To Reproduce**
Steps to reproduce the bug:
1. Go to https://move.intra.dev-toronto.ca/view/location/s1:AwnmmB/POINTS/reports/TMC ;
2. Click "Intersection Summary Report".

At this point, the total and major volumes are the same, and the table shows that there are zero minor approaches.

**Screenshots**
Intersection report header at Chesswood and Champagne:

Chesswood and Champagne with Intersection Summary report

Turning movements look reasonable, but this shows no approaches along the minor road (which should be Champagne):

table of turning movements at Chesswood and Champagne

**Additional Notes**
It's possible that we're counting the wrong leg for minor street approaches here - if the directions are flipped, for instance, we'd be counting eastbound approaches, which wouldn't exist here (since there is no road segment entering the intersection going eastbound!)

candu:
# Issue Addressed
This PR closes #666 .

# Description
This solves a bug in `ReportBaseFlowDirectional` arising from confusion between two different ways of describing the direction of a road segment from an intersection:

- "X-bound": in terms of which direction cars are travelling as they enter the intersection along that road segment;
- "X-ern": in terms of which direction that road segment lies from the intersection.

These lead to opposing directions!  For example, the _southern_ leg of an intersection contains vehicles that are travelling _northbound_ when they enter the intersection.

Our raw TMC data uses the "X-ern" interpretation, so that `S_BUS_L` means "number of busses turning left from the southern leg of the intersection".  However, our reports use the "X-bound" interpretation: this same data would fall under the "Northbound" section of a report.

`GeometryUtils.getDirectionCandidatesFrom` returns the "X-ern" direction, but we need the "X-bound" direction to correctly build the report - so this fix corrects for that in `ReportBaseFlowDirectional.getRoadDirections`.

# Tests
Created a new side-by-side test case using the original FLOW report, and used that to verify the fix.  Also verified manually on a 3-way intersection in the dev dataset (Rosemount + Greenlaw).  Used pre-commit checks to ensure that fix causes no regressions in other reports subclassing `ReportBaseFlowDirectional`.


candu (commented):
LGTM

PR #668, opened 2020-10-15
Midblock descriptions: web integration

### LINKED ISSUE #539: Beta prep: better descriptions for midblocks, corridors
**Description**
As part of several multi-location user flows, we need to be able to clearly describe individual selected features, including any such features that are part of a routed corridor.  Both corridor and non-corridor selections can include midblocks, which right now are described using only the name of the street they are a part of.

This task seeks to provide a clear yet concise description of exact location for midblocks.

**Acceptance Criteria**
- [x] add a new materialized view `location_search.centreline_midblock` that contains "X between Y and Z" textual descriptions of midblocks;
- [ ] in `CentrelineDAO`, join against this view when fetching segments and use that for the `description` field.

**Additional Notes**
Figma designs use relative positioning to describe locations in terms of the last waypoint in a selection, as in "Don Mills Rd, 400 m North of Overlea Blvd":

image

This raises the question of midblock _waypoints_: it seems weird to give them a relative positioning description, but "Don Mills Rd" isn't enough to be clear.

Another option is endpoint naming, e.g. "Don Mills Rd, St Dennis Dr to Rochefort Dr".  This could also take into account the routed direction of a corridor, so that you'd get either "Don Mills Rd, St Dennis Dr to Rochefort Dr" or "Don Mills Rd, Rochefort Dr to St Dennis Dr" depending on that direction (with one being chosen as a default for waypoints).  The issue here is that some midblocks connect minor features, which would lead to descriptions like "Jane St, Alliance Rd to Black Creek Trl / ...".

candu:
# Issue Addressed
This PR closes #539 .

# Description
We integrate the new `location_search.centreline` materialized view into `web`.

# Tests
Added `MidblockDescription.spec.js` to test `MidblockDescription`.  Ran database tests again with new `dev_data` version.  Tested quickly in frontend.

### DIFF
--- a/web/components/FcPaneMapPopup.vue
+++ b/web/components/FcPaneMapPopup.vue
@@ -120,11 +120,14 @@ async function getCentrelineDetails(feature, centrelineType) {
 function getCentrelineDescription(feature, { location }) {
   const description = [];
 
-  let { name } = feature.properties;
-  if (name) {
-    name = formatCountLocationDescription(name);
+  let name = location.description;
+  if (name === null) {
+    name = feature.properties.name;
+    if (name !== null) {
+      name = formatCountLocationDescription(name);
+    }
   }
-  description.push(name);
+  description.push(location.description);
 
   const locationFeatureType = getLocationFeatureType(location);
   if (locationFeatureType !== null) {
--- a/web/components/FcPaneMapPopup.vue
+++ b/web/components/FcPaneMapPopup.vue
@@ -118,16 +118,7 @@ async function getCentrelineDetails(feature, centrelineType) {
 }
 
 function getCentrelineDescription(feature, { location }) {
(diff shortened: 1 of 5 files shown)


web/components/FcPaneMapPopup.vue
```diff
@@ -120,11 +120,14 @@ async function getCentrelineDetails(feature, centrelineType) {
 function getCentrelineDescription(feature, { location }) {
   const description = [];
 
-  let { name } = feature.properties;
-  if (name) {
-    name = formatCountLocationDescription(name);
+  let name = location.description;
```
candu, line 123: why do we do any of this if we're not using `name`?  get rid of this part.

candu (commented):
one small change, otherwise good

Issue #639, opened 2020-09-24
Improve naming convention for corridors, generated downloads


candu:
**Description**
When showing multi-location selection descriptions, we currently make no _textual_ distinction between a set of points and a connected corridor.  This task changes that by applying different naming conventions in `CORRIDOR` vs. `POINTS` mode.

**Acceptance Criteria**
- [x] add a new function `getLocationsSelectionDescription` to `CentrelineUtils`:
  - [x] in `POINTS` mode, keep the "{{firstLocation}} + {{n - 1}} locations" convention from `getLocationsDescription`;
  - [x] in `CORRIDOR` mode, use `{{firstLocation}} to {{lastLocation}}" instead;
- [x] use this function when showing multi-location descriptions, instead of `getLocationsDescription`;
- [ ] add `selectionType` to job data / metadata for `GENERATE_REPORTS`:
  - [ ] migrate existing downloads to have `POINTS` for `selectionType`;
  - [ ] use that to fetch locations for reports;
  - [ ] use that to show location description in Manage Downloads;
- [ ] clean up `getLocationsDescription`.

**Additional Notes**
We previously considered making this a post-v1 task, but together with #538 it would make a _big_ difference in usability based on what we've heard in the beta test so far.

PR #678, opened 2020-10-23
Report headers and details, corridor names

### LINKED ISSUE #538: Some report details need to be easier to read / understand
**What Happened**
From testing notes by @mkewins :

> Speed percentile — 85th percentile, average speed, AM/PM peak volumes, total number of vehicles
> Missing units? Are those important?
> Seeing notes about dates being in SI format?

These make it harder for users - even domain expert users! - to read and understand some of the top-level details and summary statistics of MOVE reports.

**What Should Happen**
These issues should be addressed, and the reports should be readable and easily understood.

**To Reproduce**
Steps to reproduce the bug:
1. Go to any location with speed / volume ATRs;
2. Click "View Data";
3. In the studies table, click "View Reports" in the "Speed / Volume ATR" row;
4. Load the speed percentile report.

At this point, you should see the issues mentioned above with speed percentile details at top of the report.

**Screenshots**
Here we see that the speed percentile report is missing AM/PM peak volumes, that the mean / percentile speeds have no units, and that dates are in `MM/DD/YYYY` format (i.e. `en-US` locale default) instead of SI (ISO 8601?) format:

image

**Additional Notes**
From [this Wikipedia article](https://en.wikipedia.org/wiki/Date_and_time_notation_in_Canada) and Adrian's notes, the preferred format for all-numeric dates in Canada is `YYYY-MM-DD`, following the ISO 8601 standard.

candu:
# Issue Addressed
This PR closes #538 and makes progress on #639 .

# Description
We make several content, layout, and interaction tweaks to reports, based on a bunch of feedback over the last while around these.

# Tests
Updated unit tests and ran those (especially around report generation).  Tested report types manually in the browser.


candu (commented):
LGTM

PR #680, opened 2020-10-26
Fetch map tiles and glyphs from current web environment

### LINKED ISSUE #679: Environment crosstalk in QA: map tiles / glyphs
**What Happened**
The MOVE QA environment uses vector tiles and font glyphs from dev ETL.

**What Should Happen**
These tiles should be served from within the same environment, most likely from the webserver.

**To Reproduce**
Steps to reproduce the bug:
1. Open MOVE in the QA environment on any page that shows a map.
2. Open your browser's devtools, and go to the Network tab.
3. Search for requests matching "intra.dev-toronto.ca".

At this point, you'll see several requests for vector tiles and font glyphs from dev ETL.

**Additional Notes**
The culprit appears to be `addTippecanoeSource` from `lib/geo/GeoStyle`, which hardcodes the use of dev ETL here.

We do seem to have tiles / glyphs `nginx` configs in web, so there shouldn't be any required backend work.

candu:
# Issue Addressed
This PR closes #679 .

# Description
We stop fetching these from DEV0 ETL, and start fetching them from the appropriate web environment.

# Tests
Ran coverage tests, including `GeoStyle.spec.js`.

### DIFF
--- a/lib/geo/GeoStyle.js
+++ b/lib/geo/GeoStyle.js
@@ -4,6 +4,20 @@ import metadataDark from '@/lib/geo/theme/dark/metadata.json';
 import rootStyleLight from '@/lib/geo/theme/light/root.json';
 import metadataLight from '@/lib/geo/theme/light/metadata.json';
 
+// MAP ASSETS ORIGIN
+
+/**
+ * @returns {string} origin for map tiles and glyphs
+ */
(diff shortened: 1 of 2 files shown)


candu (commented):
LGTM

Issue #688, opened 2020-10-30
Pentest preparation tasks


candu:
**Description**
This is a tracking issue to capture some of the technical work needed to get to pentest readiness.

**Acceptance Criteria**
- [x] mock out email sending in local environments;
- [x] remove SendGrid from dependencies;
- [x] remove `autoindex` and `Access-Control-Allow-Origin` header from `/tiles` and `/glyphs`;
- [x] create DAG to generate `/glyphs` in new environments;
- [x] review MOVE dependencies to see which ones are essential;
- [x] coordinate around process changes:
  - [x] CD from `deploy-qa` branch in QA, rather than `master`
  - [x] check sandbox stack, ensure user stories work
- [x] configure `Content-Security-Policy` header in `nginx`;
- [x] deploy new version to QA and:
  - [x] ensure above fixes / changes;
  - [x] ensure environment crosstalk has been solved.

candu:
Marking as closed after recent deployment 🎉

PR #690, opened 2020-11-02
Pentest prep grab bag: CSP header, dependency audit, cleaning up nginx configs

candu:
# Issue Addressed
This PR makes significant progress on #688 .

# Description
We clean up configs and dependencies, and add support for the `Content-Security-Policy` header at the `nginx` level of the stack.  In `nginx` configs, we use the `$request_id` to generate a nonce, which is then passed via the `<meta name="csp-nonce">` tag to Vuetify to configure its dynamically-generated theme tags.

# Tests
Tested parts of this locally, but the only way to properly test it is by deploying to DEV0 and checking that our CSP setup works as expected.


vue.config.js
```diff
@@ -15,6 +15,21 @@ const cert = fs.readFileSync(path.join(__dirname, 'ssl', 'localhost.crt'));
...
+      /*
+       * This header is slightly different from that used in our production `nginx` configs
+       * in a few ways:
+       *
+       * - we allow `move.intra.dev-toronto.ca` as a `connect-src` to load map assets
+       *   from the dev environment in development, but load these assets from `'self'` in
+       *   production;
+       * - we allow `'unsafe-eval'` as a `script-src` in development, but load scripts from
+       *   `self` in production;
+       * - we allow `'unsafe-inline'` as a `style-src-elem` in development, but use a nonce in
```
candu, line 28: s/style-src-elem/style-src/

candu (commented):
mostly good, one tiny docs change

PR #693, opened 2020-11-03
Caching, CSP, security headers for nginx configuration

candu:
# Issue Addressed
This PR makes further progress on #688 .

# Description
Fixing CSP issues around location matching, adding caching, adding X-Frame-Options and X-Content-Type-Options.

# Tests
Copied config changes to dev environment, restarted `nginx`, verified that:

- configuration is valid (i.e. no error messages in `/var/log/nginx/error.log`);
- headers appear as expected;
- caching works as expected (by unchecking "disable cache" in devtools, reloading page, observing transferred vs. resources size).

Will complete testing after deployment to dev.

### DIFF
--- a/scripts/deployment/web/nginx/default.d/flashcrow.conf
+++ b/scripts/deployment/web/nginx/default.d/flashcrow.conf
@@ -1,3 +1,6 @@
+#
+# MOVE backend services
+#
 location /api {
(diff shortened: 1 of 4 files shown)


candu (commented):
LGTM

Issue #696, opened 2020-11-03
More configuration updates in preparation for pentest


candu:
**Description**
More configuration changes requested in preparation for setting these environments up.

**Acceptance Criteria**
- [x] remove `nginx`, `postgres` lines from `install.sh`;
- [x] give OK to remove `shellcheck`;
- [x] set Same-Site cookie attribute to "Strict", rather than "Lax";
- [x] remove HSTS, other X-headers from `nginx` / `application`;
- [x] give OK to remove SendGrid credentials from SSM parameter;
- [x] advise re:
  - [x] DB schema updates: corruption if two instances are being deployed at the same time?
  - [x] `scheduler` handling multiple AWS instances?
- [x] remove `nvm` from configs;
- [x] switch over to unprivileged app-specific service user;
- [x] deploy and test changes.

candu:
Marking as closed for now - if there are additional small changes, those will likely be merged directly to master.

PR #697, opened 2020-11-04
Remove unneeded dependencies, security headers

candu:
# Issue Addressed
This PR makes progress on #696 .

# Description
Following up on discussion with Cloud Services - many of these should be handled at the CloudFormation / AWS level.

# Tests
Tested Same-Site changes quickly; need to deploy to dev to test other changes.


candu (commented):
LGTM

PR #699, opened 2020-11-05
Update CodeDeploy scripts: remove nvm, switch to app user

candu:
# Issue Addressed
This PR makes further progress on #696 .

# Description
We remove `nvm` to rely instead on pinned LTS versions of `node` and `npm`, and switch over to using an unprivileged app-specific service user.

# Tests
Will test by deploying to dev (and possibly follow up with a couple of quick config changes on `master` before going to QA).

### DIFF
--- a/deploy_scripts/install.sh
+++ b/deploy_scripts/install.sh
@@ -1,31 +1,5 @@
 #!/bin/bash
-# shellcheck disable=SC1090,SC1091
 
 set -euo pipefail
 
-set +u
-source /home/ec2-user/.bash_profile
-set -u
-
-# section: pre_install_flashcrow
-## /install_node.sh
-if command -v nvm > /dev/null 2>&1; then
-  echo "nvm already installed, skipping..."
-else
-  echo "installing nvm..."
-  curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.36.0/install.sh | bash
-  # ensure that nvm shims are available in current shell session
-  export NVM_DIR="$HOME/.nvm"
-  [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
-  [ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
-fi
-
-# confirm
-echo "$NVM_DIR"
-
-# install correct version of node
-echo "installing node@lts/*..."
-nvm install lts/*
-echo "lts/*" > /home/ec2-user/.nvmrc
-nvm use lts/*
 npm install -g forever
--- a/deploy_scripts/install.sh
+++ b/deploy_scripts/install.sh
@@ -2,4 +2,4 @@
 
 set -euo pipefail
 
-npm install -g forever
+echo "Nothing to install."
--- a/deploy_scripts/start.sh
+++ b/deploy_scripts/start.sh
@@ -3,24 +3,25 @@
 
 set -euo pipefail
 
-set +u
-source /home/ec2-user/.bash_profile
-set -u
+# copy repo to /var/app/flashcrow
+sudo rm -rf /var/app/flashcrow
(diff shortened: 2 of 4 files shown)


deploy_scripts/install.sh
```diff
@@ -1,31 +1,5 @@
...
-
-# confirm
-echo "$NVM_DIR"
-
-# install correct version of node
-echo "installing node@lts/*..."
-nvm install lts/*
-echo "lts/*" > /home/ec2-user/.nvmrc
-nvm use lts/*
 npm install -g forever
```
candu, line 5: this won't work, as there's no global `node` to call - we'd need the `PATH` that's set in `start.sh` to make this work.

instead of doing that, though, let's just move this down to `start.sh` - it's a bit strange in that we're then installing dependencies in the start script, but it does allow us to avoid the complexity of setting this `PATH` in two separate scripts.

deploy_scripts/start.sh
```diff
@@ -3,24 +3,25 @@
...
 
-# copy nginx / forever configs from repo
+# copy nginx configs to /etc/nginx
 sudo cp /home/ec2-user/flashcrow/scripts/deployment/web/nginx/nginx.conf /etc/nginx/
 sudo cp /home/ec2-user/flashcrow/scripts/deployment/web/nginx/default.d/*.conf /etc/nginx/default.d/
-cp /home/ec2-user/flashcrow/scripts/deployment/web/forever.json /home/ec2-user/forever.json
 
 # make log directory
-mkdir -p /home/ec2-user/log/flashcrow
+sudo mkdir -p /var/app/log
```
candu, line 20: need to also change `forever.json` in this PR, which we aren't doing yet!

deploy_scripts/start.sh
```diff
@@ -3,24 +3,25 @@
...
+# copy repo to /var/app/flashcrow
+sudo rm -rf /var/app/flashcrow
+sudo cp -R ~ec2-user/flashcrow /var/app
 
-# copy private config to repo
-cp -r /home/ec2-user/flashcrow.config.js /home/ec2-user/flashcrow/lib/config/private.js
+# copy private / forever configs to /var/app/flashcrow
+sudo chown -R ec2-user:ec2-user /var/app/flashcrow
+cp -r /home/ec2-user/flashcrow.config.js /var/app/flashcrow/lib/config/private.js
+cp /var/app/flashcrow/scripts/deployment/web/forever.json /var/app/flashcrow/forever.json
```
candu, line 13: there's actually no reason to do this - we can simplify by just passing `/var/app/flashcrow/scripts/deployment/web/forever.json` as the `forever` config below.

deploy_scripts/start.sh
```diff
@@ -34,11 +35,12 @@ sudo cp -r /home/ec2-user/flashcrow/dist /usr/share/nginx/html/flashcrow
...
 . /home/ec2-user/psqlArgs.config
 # shellcheck disable=SC2154
-/home/ec2-user/flashcrow/scripts/db/db-update.sh --psqlArgs "$psqlArgs"
+/var/app/flashcrow/scripts/db/db-update.sh --psqlArgs "$psqlArgs"
 
 # start flashcrow
+cd /var/app/flashcrow
 # shellcheck disable=SC2046
-env $(xargs < /home/ec2-user/cot-env.config) NODE_ENV=production NODE_EXTRA_CA_CERTS=/home/ec2-user/flashcrow/ssl/extra-ca-certs.cer forever start /home/ec2-user/forever.json
+sudo -u appsvc env $(xargs < /var/app/config/cot-env.config) PATH="$PATH" NODE_ENV=production NODE_EXTRA_CA_CERTS=/var/app/ssl/extra-ca-certs.cer forever start /var/app/flashcrow/forever.json
```
candu, line 43: here: we can set `forever`'s config file via that last argument, which means we don't need to re-copy `forever.json` at all!

candu (commented):
several changes requested

PR #700, opened 2020-11-10
Add "tcs:" as special term for signal search

### LINKED ISSUE #683: Signal search doesn't include "tcs:" terminology
**Description**
Although we have signal search in MOVE, users aren't clear on how to use it.  One easy step here is to add "tcs:" as another way to search by signal, so that we're at least using current terminology.

**Acceptance Criteria**
- [x] users can type `tcs:` followed by the signal ID to search by signal.

**Additional Notes**
This doesn't help solve the discoverability problem, but it should at least help in terms of language.

candu:
# Issue Addressed
This PR closes #683 .

# Description
We add "tcs:" alongside "px:" and "signal:" as a special search term for traffic control signals.

# Tests
Updated `LocationSearchDAO.spec.js` and ran tests.


candu (commented):
LGTM

PR #701, opened 2020-11-10
Time range filters for View Data

### LINKED ISSUE #698: Date range support for View Data filters: quick fix
**Description**
One key feature gap identified through the longitudinal beta is _date range filtering_: users would like to be able to filter by date range.  A full exploration and implementation would take longer than we have pre-launch, but we do have time to make a couple of small changes that should make this a bit more usable.

**Acceptance Criteria**
- [ ] in View Data filters, add option to filter by a specific date range...
  - [ ] collision filters
  - [ ] study filters
- [ ] in map legend, add a hint below the dropdown that shows active date range
- [ ] if possible: offer "all" (or "most recent") as an option in map legend dropdown

**Additional Notes**
We don't likely have time to do a full implementation of "all" - what I'd suggest there is handling it just for dynamic layers, and leaving the pre-built collision data layers alone.

candu:
# Issue Addressed
This PR closes #698 .

# Description
We replace `datesFrom` in our collision and study filters with three filter parameters: `applyDateRange`, which is used to flag whether a date range is being applied, and `dateRangeStart` / `dateRangeEnd` for the date range itself.  We then update the filter dialog UI and surrounding logic to match.

# Tests
Tested different filter groups in frontend to ensure that filter UI works correctly, that chips are shown correctly, that GET params are sent correctly, and that clicking filters removes them correctly.

### DIFF
--- a/lib/validation/ValidationsFilters.js
+++ b/lib/validation/ValidationsFilters.js
@@ -1,0 +1,12 @@
+import { requiredIf } from 'vuelidate/lib/validators';
+
+const ValidationsFilters = {
+  dateRangeStart: {
+    requiredIfApplyDateRange: requiredIf(({ applyDateRange }) => applyDateRange),
+  },
+  dateRangeEnd: {
+    requiredIfApplyDateRange: requiredIf(({ applyDateRange }) => applyDateRange),
+  },
+};
+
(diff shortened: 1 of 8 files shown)


candu (commented):
LGTM

PR #704, opened 2020-11-16
Use Content-Disposition header to set filename in backend

candu:
# Issue Addressed
This PR makes further progress on #639 .

# Description
We centralize the generation of filenames in `StoragePath`, and return the filename via the `Content-Disposition` header.  We then refactor the frontend to use that header by dynamically creating and "clicking" an `<a href="..." download>` link.  This allows us to remove `file-saver` from 2 of 3 callsites (it's still used in Track Requests to download requests).

In the process, we also fix a bug where a `curl -I` request for a CSV report would crash MOVE Reporter - this turned out to be a small issue with how `csv-stringify` handles streams.

# Tests
Tested quickly in local development: CSV and PDF reports both work; this change doesn't break bulk report generation; these reports can be fetched via `curl`; `curl -I` requests work properly for all report formats.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -353,24 +353,30 @@ ReportExportMode.init({
 /**
  * Report formats available from MOVE Reporter.
  *
+ * @param {boolean} download - whether reports of this format should be downloaded; used to
+ * set `Content-Disposition` header.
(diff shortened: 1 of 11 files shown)


candu (commented):
LGTM

PR #705, opened 2020-11-17
User-friendly report naming

### LINKED ISSUE #639: Improve naming convention for corridors, generated downloads

candu:
# Issue Addressed
This PR closes #639 .

# Description
We refactor `StoragePath` considerably, making the `forXXX` methods async so that we can fetch information from database to build filenames here.  To help with that, we add `selectionType` to `GENERATE_REPORTS` jobs in `scheduler`, and update the whole report stack to use this plus `RoutingDAO` (via the new `FeatureResolver`) instead of a pre-resolved set of locations.

# Tests
Ran `npm run ci:jest-coverage` to test, which includes revamped `StoragePath` testing.  Generated several reports, both single and bulk, and both web and download, to ensure this doesn't break anything.

### DIFF
--- a/lib/controller/LocationController.js
+++ b/lib/controller/LocationController.js
@@ -4,12 +4,14 @@ import {
   CentrelineType,
   FEATURE_CODES_INTERSECTION,
   FEATURE_CODES_SEGMENT,
+  HttpStatus,
+  LocationSelectionType,
   LocationSearchType,
-  MAX_LOCATIONS,
 } from '@/lib/Constants';
 import CentrelineDAO from '@/lib/db/CentrelineDAO';
 import LocationSearchDAO from '@/lib/db/LocationSearchDAO';
-import RoutingDAO from '@/lib/db/RoutingDAO';
+import { InvalidFeaturesSelectionError } from '@/lib/error/MoveErrors';
+import FeatureResolver from '@/lib/geo/FeatureResolver';
 import CompositeId from '@/lib/io/CompositeId';
 import Joi from '@/lib/model/Joi';
 import CentrelineLocation from '@/lib/model/helpers/CentrelineLocation';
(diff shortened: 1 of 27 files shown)


lib/controller/LocationController.js
```diff
@@ -203,20 +205,21 @@ LocationController.push({
...
+      features,
+      selectionType: LocationSelectionType.CORRIDOR,
+    };
+    try {
+      const featuresResolved = await FeatureResolver.byFeaturesSelection(featuresSelection);
+      const locations = await CentrelineDAO.byFeatures(featuresResolved);
+      return filterLocations(locations);
+    } catch (err) {
+      if (err instanceof InvalidFeaturesSelectionError) {
+        const { statusCode } = HttpStatus.BAD_REQUEST;
```
candu, line 218: as we found with the test, this _does_ mean that "corridor not found" results in HTTP 400, not HTTP 404.  (this isn't urgent or even necessary to fix, especially while we don't have any status code-specific handling in the frontend!)

lib/io/storage/StoragePath.js
```diff
@@ -13,55 +27,260 @@ import ObjectUtils from '@/lib/ObjectUtils';
...
+    return `${storagePaths.length}_TOTAL`;
+  }
+
+  /**
+   *
+   * @param {string} description
+   * @returns {string}
+   */
+  static sanitizeLocationDescription(description) {
+    return description
```
candu, line 156: this doesn't currently perform any truncation!  we should ensure that it does, and add test cases for this in `StoragePath.spec.js`.

lib/io/storage/StoragePath.js
```diff
@@ -1,6 +1,20 @@
 import crypto from 'crypto';
```
candu, line 1: this file feels like the kind of thing that really, really, really needs some kind of fuzz testing.  (same with our reports - we should be making database-backed tests for this!)

lib/reports/ReportBaseCrash.js
```diff
@@ -51,9 +40,17 @@ class ReportBaseCrash extends ReportBase {
...
+    let info = getLocationsSelectionDescription(locationsSelection);
+    /*
+     * `pdfmake` only supports ANSI characters:
+     *
+     * https://pdfmake.github.io/docs/0.1/fonts/standard-14-fonts/
+     *
+     * To work around this, we replace the Unicode arrow used in corridor descriptions
+     * with a more ANSI-friendly equivalent here.
+     */
+    info = info.replace('\u2192', '-');
```
candu, line 53: we should also handle `\u2013` here, which is generated by `MidblockDescription` for midblocks with two detectable cross streets.

lib/reports/ReportIdParser.js
```diff
@@ -0,0 +1,129 @@
+/* eslint-disable class-methods-use-this */
```
candu, line 1: can we add a test for this?

candu (commented):
a few small changes, but mostly good

PR #712, opened 2020-11-24
Reduced session timeouts

### LINKED ISSUE #711: Reduce timeout for ADMIN users
**Description**
There are business reasons to allow regular users to remain logged in for a full workday - with legacy tools, users often returned to them periodically throughout the day, meaning that inactive timeouts would force them to log in again frequently.  User research and testing strongly suggests that MOVE usage will follow a similar pattern.

That said, it does make sense to limit the duration of `ADMIN` sessions, as these carry greater risk if left open unattended.

**Acceptance Criteria**
- [x] sessions for `ADMIN` users timeout faster (either with a shorter duration, or after a short period of inactivity);
- [x] sessions for non-`ADMIN` users remain valid as before.

**Additional Notes**
The session timeout is currently set to 1 day in `MoveServer` via the `ttl` setting.  From [cookie auth plugin docs](https://hapi.dev/module/cookie/api/?v=11.0.2), it seems that `keepAlive` controls whether this is inactive time (if `true`) or total time (if `false`).

This is, however, global across all users.  There may be a way to set this for a specific session using the `request.cookieAuth.ttl()` method, but we'll need to figure out where in the request lifecycle to call that.  (In `validateFunc`?  In the `onPostAuth` hook?  Or maybe even in the `AdminController` endpoints?)

candu:
# Issue Addressed
This PR closes #711 .

# Description
We reduce the timeout for admin users to 5 minutes, and for non-admin users to 9 hours.  We also add `keepAlive: true` to our cookie authentication settings, so that these timeouts occur after the specified period of _inactivity_.

Since our CI tests don't actually set the authentication cookie, we also also disable this timeout setting when `config.ENV === 'test'`.

# Tests
Tested that users can still log in and out.  Tested that a non-admin user is given a session good for 9 hours, and that this period is renewed on each request while logged in.  Tested that an admin user is given a session good for 5 minutes, and that this period is renewed on each request while logged in.  Tested directly that leaving an admin session open for 5 minutes with no activity results in the session expiring.  Ran all unit, database, and REST API tests in CI to double-check that this doesn't break anything.

### DIFF
--- a/lib/auth/SessionTimeouts.js
+++ b/lib/auth/SessionTimeouts.js
@@ -1,0 +1,20 @@
+export default {
+  /*
+   * For admin users, we override the default TTL with a much shorter duration.  This
+   * helps mitigate the risk of leaving an admin session open unattended.
+   *
+   * See the `onPostAuth` hook in `MoveServer`, and `login` in `AuthController`,
+   * for callsites.
+   */
+  TTL_ADMIN: 5 * 60 * 1000,
+
+  /*
+   * Default TTL for user sessions.
+   *
+   * This is deliberately set to the length of core business hours (8am-5pm) at the City
+   * of Toronto.  Sessions started by non-admin users at start-of-day are guaranteed to
+   * remain open through end-of-day, and are also guaranteed to expire before
+   * start-of-next-day.
+   */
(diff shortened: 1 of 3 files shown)


candu (commented):
LGTM

Issue #715, opened 2020-11-30
Intersection summary and TCS reports failing at Bloor and High Park 


candu:
**What Happened**
On [View Reports at Bloor and High Park](https://move.intra.dev-toronto.ca/view/location/s1:A4WvmB/POINTS/reports/TMC), the TMC Summary and TMC Detailed reports work, but the Intersection Summary and warrant reports do not.

**What Should Happen**
All four report types should work at all intersections.

**To Reproduce**
Steps to reproduce the bug:
1. Go to [View Reports at Bloor and High Park](https://move.intra.dev-toronto.ca/view/location/s1:A4WvmB/POINTS/reports/TMC)
2. Click on the "Intersection Summary Report" tab near top.

At this point, the loading spinner remains indefinitely, and the report does not load.

**Screenshots**
Loading spinner:

![loading spinner for Intersection Summary report at Bloor and High Park](https://user-images.githubusercontent.com/151605/100618228-8b411880-32e9-11eb-982a-3c9b950bad49.png)

Stacktrace from `reporter` for intersection summary report:

```
Debug: handler, error
    TypeError: Cannot read property 'short' of null
    at Function.computeHourlyTotals (/var/app/flashcrow/lib/reports/ReportIntersectionSummary.js:80:56)
    at /var/app/flashcrow/lib/reports/ReportIntersectionSummary.js:162:50
    at Array.map (<anonymous>)
    at ReportIntersectionSummary.transformData (/var/app/flashcrow/lib/reports/ReportIntersectionSummary.js:161:37)
    at ReportIntersectionSummary.generate (/var/app/flashcrow/lib/reports/ReportBase.js:126:34)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async handler (/var/app/flashcrow/lib/controller/ReportController.js:121:26)
    at async exports.Manager.execute (/var/app/flashcrow/node_modules/@hapi/hapi/lib/toolkit.js:60:28)
    at async Object.internals.handler (/var/app/flashcrow/node_modules/@hapi/hapi/lib/handler.js:46:20)
    at async exports.execute (/var/app/flashcrow/node_modules/@hapi/hapi/lib/handler.js:31:20)
```

The warrant report stacktrace is similar, with the same `TypeError` in `computeHourlyTotals`.

**Additional Notes**
[...]

PR #716, opened 2020-11-30
Allow for floating-point approximate matches in GeometryUtils

candu:
# Issue Addressed
This PR makes progress on #715 .

# Description
We account for a recent update to coordinate precision to `gis.centreline` (but not `gis.centreline_intersection`!) by introducing `DEG_EPSILON` in `GeometryUtils`.

# Tests
Added a test case based on a specific location (Don Mills & Overlea!) where this issue occurred.  Ran rest of `GeometryUtils` tests, and all the dependent tests, during pre-commit; running all CI tests.

Based on the investigation, this is extremely likely to fix the issue, but we will test manually in AWS dev to confirm.


candu (commented):
LGTM

PR #731, opened 2020-12-14
Use production DCS ID in production environment

### LINKED ISSUE #728: Production analytics
**Description**
MOVE is sending analytics events in production, but we've been unable to see these in the Oracle Infinity dashboard.  In addition, these events don't contain browser type / version info, making it hard to focus our compatibility efforts there.

**Acceptance Criteria**
- [x] switch to Oracle Infinity production tag in production;
- [x] verify that production domain is added to Oracle Infinity configuration.

candu:
# Issue Addressed
This PR closes #728 .

# Description
We use the analytics domain to determine which DCS ID to load for Oracle Infinity, based on the values in our tags.

# Tests
Will need to test production configuration in production, but non-production configuration works so far.  (The worst possible impact from this would be that we continue not having analytics in production - not ideal, but also not worse than present.)


candu (commented):
LGTM

PR #732, opened 2020-12-15
Updates to collisions, studies layers and popups

### LINKED ISSUE #725: Collision popups rendering for wrong collision
**What Happened**
As reported via MOVE:

> there is a collision noted on Lionhead Trail in Scarborough that says its a collision on Dundas Street

After a bit of exploration, this appears to be a more general problem where collision popups render for the wrong collision at certain zoom levels.

**What Should Happen**
Collision popups should always render for the right collision.

**To Reproduce**
Steps to reproduce the bug:
1. Go to a location with many collisions;
2. Zoom to about zoom level 16.5;
3. Hover over collisions in the area until you find one where the collision popup renders for the wrong collision;
4. Zoom in a bit more, past zoom level 17;
5. Roll over the same collision, which should now render the popup for the right collision.

The bug here is exhibited at step 3, but steps 4 and 5 provide additional detail that might help track this down.

**Screenshots**
A collision near Avenue and St. Clair that's misidentified as University and Gerrard:

collision at Avenue and St. Clair, misidentified as University and Gerrard

Same circle, when we zoom in, is now correctly identified as 76 Dunvegan Rd:

same collision, now identified as 76 Dunvegan Rd

Looking at the collisions requested, it seems these are indeed two different collisions with consecutive IDs:

these two collisions have consecutive IDs

**Additional Notes**
Similar "consecutive ID" behaviour is found at other locations - this suggests that there's some kind of ID reassignment going on (e.g. when it culls / clusters points that are nearby each other), but we don't have direct confirmation of that yet.

candu:
# Issue Addressed
This PR closes #725 .

# Description
Several small changes here:

- use layer filters instead of separate sources to handle 1, 3, 5, 10 year switching;
- show all locations with studies on map, and use greyed-out icon to represent studies not matching the date filter;
- building on those changes, drop parameterized layers from several places in frontend and backend.

These changes should also help prepare us for better map filtering.

# Tests
Attempted to repro #725 with changes, could not.  Manual testing in browser.


lib/geo/GeoStyle.js
```diff
@@ -189,24 +192,26 @@ normalizeAndSeparateLayers(STYLE_BASE_AERIAL);
 
 // MOVE SOURCES
 
-function addTippecanoeSource(style, id, minLevel, maxLevel, crossfade = 0) {
+function addTippecanoeSource(style, id, minLevel, maxLevel, crossfade = 0, options = {}) {
```
candu, line 195: this isn't necessary yet, strictly speaking, but it will help if we want to introduce clustering for collisions in the near future (which we well might, to help reduce confusion here...)

tests/jest/db/DynamicTileDAO.spec.js
```diff
@@ -46,21 +46,13 @@ test('DynamicTileDAO.getTileFeatures', async () => {
...
-  ).rejects.toBeInstanceOf(InvalidDynamicTileLayerError);
-
   // tile outside city boundaries
   await expect(
     DynamicTileDAO.getTileFeatures('hospitalsLevel1', 12, 345, 678),
   ).resolves.toHaveLength(0);
 
   // tile with features
-  const tileFeatures = await DynamicTileDAO.getTileFeatures('collisionsLevel1:3', 16, 18308, 23913);
+  const tileFeatures = await DynamicTileDAO.getTileFeatures('collisionsLevel1', 16, 18308, 23913);
```
candu, line 55: please also run `test:test-db` with the new dataset...

candu (commented):
run database tests please, otherwise good

Issue #695, opened 2020-11-03
Collision inconsistencies between Flow and MOVE


candu:
**What Happened**
A couple of users reported that certain collisions are listed differently in MOVE vs. CRASH, or are available in one but not the other.

**What Should Happen**
Collisions should be consistently available between MOVE and CRASH.

**To Reproduce**
Steps to reproduce the bug:
1. Go to MOVE;
2. Search for "Cummer and Snowcrest", select intersection;
3. Pan slightly east to see the KSI just on the west side of the Don;
4. Click the nearest midblock and View Data.

This collision is visible on the map.  On CRASH, it's associated with this location, whereas on MOVE it does not show up in the midblock KSI statistics.

**Screenshots**
KSI shown on map, but not in View Data:

![KSI visible on map near Cummer and Snowcrest, but not shown in KSI statistics for nearest midblock](https://user-images.githubusercontent.com/151605/98010504-c886b900-1dc4-11eb-9d49-d359c47f5494.png)

**Additional Notes**
We can see how this collision is conflated with the centreline:

```sql
flashcrow=> select * from collisions.events_centreline where collision_id = 1639231;
 collision_id | centreline_type | centreline_id
--------------+-----------------+---------------
      1639231 |               2 |      30005164
(1 row)
```

This location is actually a psuedo-intersection with the Don Trail:

```sql
flashcrow=> select * from gis.centreline_intersection where int_id = 30005164;
   gid   |  int_id  | elev_id |            intersec5            | classifi6 |               classifi7                | num_elev | elevatio9 | elevatio10 | elev_level | elevation | elevatio13 | height_r14 | height_r15 |     x      |      y      |   longitude   |   latitude   | objectid |                        geom
---------+----------+---------+---------------------------------+-----------+----------------------------------------+----------+-----------+------------+------------+-----------+------------+------------+------------+------------+-------------+---------------+--------------+----------+----------------------------------------------------
 2122383 | 30005164 |   44792 | Cummer Ave / East Don River Trl | SEUML     | Pseudo Intersection-Overpass/Underpass |        2 |    509200 | Pseudo     |          0 |         0 | F          |          0 | F          | 314387.588 | 4850403.116 | -79.380665419 | 43.795527833 |    87436 | 0101000020E6100000AE547DD25CD853C079E524DBD3E54540
[...]

PR #736, opened 2020-12-22
Update MOVE web backend to use new centreline conflation target

candu:
# Issue Addressed
This PR makes progress on #695 .

# Description
We update the web application backend to use our new centreline conflation target and routing target, rather than raw joins using `gis`, `volume`, and `location_search`.  This should provide a performance benefit by reducing joins in `CentrelineDAO`, which is used in several places throughout MOVE.

# Tests
Tested quickly in frontend.

### DIFF
--- a/tests/jest/db/CentrelineDAO.spec.js
+++ b/tests/jest/db/CentrelineDAO.spec.js
@@ -9,10 +9,10 @@ afterAll(() => {
 
 test('CentrelineDAO.byFeature', async () => {
   // intersection
-  let feature = { centrelineId: 30000549, centrelineType: CentrelineType.INTERSECTION };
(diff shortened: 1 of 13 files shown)


tests/jest/db/CentrelineDAO.spec.js
```diff
@@ -9,10 +9,10 @@ afterAll(() => {
 
 test('CentrelineDAO.byFeature', async () => {
   // intersection
-  let feature = { centrelineId: 30000549, centrelineType: CentrelineType.INTERSECTION };
```
candu, line 12: did we change this in our REST API tests as well?

candu: done.

candu (commented):
LGTM

PR #737, opened 2020-12-23
Update LocationSearchDAO to use new conflation target-backed views

candu:
# Issue Addressed
This PR makes progress on #695 .

# Description
We update `LocationSearchDAO` to use new conflation target-backed views here.

# Tests
CI tests; tested briefly in browser.


candu (commented):
LGTM

PR #749, opened 2021-01-19
CSV formats for collision reports

### LINKED ISSUE #743: CSV format for Collision Tabulation report
**Description**
During one of the MOVE training sessions, we came across an issue where CSV download of Collision Tabulation reports does not work - this is because the CSV format is not currently implemented for this report type.

This task implements the CSV format for the Collision Tabulation report.

**Acceptance Criteria**
- [ ] add a method `ReportCollisionTabulation#generateCsv` to override the version in `ReportBase`;
- [ ] from that method, return tabular data with the following columns:
  - dim1
  - dim2
  - dim1Value
  - dim2Value
  - type
  - n
- [ ] include all individual cross-tabulations from the report in this dataset;
- [ ] for `type`, use either `EVENT` or `INVOLVED`, based on the value of the `event` flags for the two dimensions;
- [ ] for hour-of-day and day-of-week, set `dim2` and `dim2Value` to `null`, and only use `dim1` / `dim1Value` / `numCollisions`.

candu:
# Issue Addressed
This PR closes #743 .

# Description
We implement the CSV download format for tabulation reports, and revamp the CSV format for collision directory reports to match the layout of web / PDF reports there.

# Tests
Tested quickly in frontend.

### DIFF
--- a/lib/reports/ReportCollisionDirectory.js
+++ b/lib/reports/ReportCollisionDirectory.js
@@ -69,21 +69,112 @@ class ReportCollisionDirectory extends ReportBaseCrash {
      * TODO: this violates the "no shared references" invariant in `ReportBase#transformData`.
      * We should address that eventually, but there isn't time right now.
      */
-    const rawCollisions = rawData.collisions;
-    const collisions = rawCollisions.map(
+    const collisions = rawData.collisions.map(
       ReportCollisionDirectory.getCollisionDirectoryEntry,
     );
     const collisionSummary = { ...rawData.collisionSummary };
-    return { collisions, collisionSummary, rawCollisions };
+    return { collisions, collisionSummary };
   }
 
-  generateCsv(location, { rawCollisions }) {
-    if (rawCollisions.length === 0) {
-      return { columns: [], rows: [] };
-    }
-    const columns = Object.keys(rawCollisions[0])
-      .map(key => ({ key, header: key }));
-    return { columns, rows: rawCollisions };
+  getCsvDriverCells({
+    drivact = null,
+    drivcond = null,
+    initdir = null,
+    invage = null,
+    manoeuver = null,
+  }) {
+    return {
+      drivact: this.getCollisionFactorCode('drivact', drivact),
+      drivcond: this.getCollisionFactorCode('drivcond', drivcond),
+      initdir: this.getCollisionFactorCode('initdir', initdir),
+      invage: this.getCollisionFactorCode('invage', invage),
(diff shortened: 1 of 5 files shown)


lib/reports/ReportCollisionDirectory.js
```diff
@@ -69,21 +69,112 @@ class ReportCollisionDirectory extends ReportBaseCrash {
      * TODO: this violates the "no shared references" invariant in `ReportBase#transformData`.
```
candu, line 69: get rid of this comment - now that we've removed `rawCollisions` here, it's no longer true!

lib/reports/ReportCollisionDirectory.js
```diff
@@ -69,21 +69,112 @@ class ReportCollisionDirectory extends ReportBaseCrash {
...
   }
 
-  generateCsv(location, { rawCollisions }) {
-    if (rawCollisions.length === 0) {
-      return { columns: [], rows: [] };
-    }
-    const columns = Object.keys(rawCollisions[0])
-      .map(key => ({ key, header: key }));
-    return { columns, rows: rawCollisions };
+  getCsvDriverCells({
```
candu, line 79: right now, we have no tests on these reports at all.  since these access `collision_factors`, we'd either have to a) put them in our database-backed tests, or b) mock out `collision_factors` fetching to make these unit tests.

another possible approach is that collision reports - especially tabulation reports! - have a bunch of logic that could be unit tested independently of the full end-to-end pipeline.

lib/reports/ReportCollisionTabulation.js
```diff
@@ -331,6 +331,119 @@ class ReportCollisionTabulation extends ReportBaseCrash {
...
+      let valueY;
+      if (Array.isArray(y)) {
+        valueY = dimY.otherDescriptionFn(y);
+      } else {
+        valueY = dimY.entries.get(y).description;
+      }
+
+      orderedTable[i].forEach((value, j) => {
+        if (j >= nx) {
+          return;
```
candu, line 352: add a comment here to explain this

candu (commented):
needs more testing, otherwise good.

lib/reports/ReportCollisionDirectory.js
candu: refactored this a bit to make unit testing of that separable logic easier in the future, but we should tackle this when we have more time.

candu (commented):
LGTM

PR #751, opened 2021-01-19
Fix Client, Study Type columns in requests CSV export

### LINKED ISSUE #744: Missing fields and other issues with study requests CSV download
**What Happened**
From a user-submitted bug report:

> The issue is related to the download function. There are some fields missing (e.g. Study Type), while others are a bit different than expected (e.g. the requester field has the "ORG/" prefix. The default CSV encoding is also UTF-8 I believe, which creates issues when using in Excel (without additional steps) for certain characters.

**What Should Happen**
Study requests CSV download should contain valid Study Type values and strip the "ORG/" prefix from usernames.  (I don't _think_ we have non-ASCII characters in these exports, but possibly worth checking?)

**To Reproduce**
Steps to reproduce the bug:
1. Log into MOVE.
2. Click on "Track Requests" in the left navbar.
3. Select all requests from the table.
4. Click "Download".
5. Open the resulting CSV.

At this point, you should see the CSV, but the `studyType` column is empty.

candu:
# Issue Addressed
This PR closes #744 .

# Description
Fixes to a couple of columns in the Track Requests CSV export.

# Tests
Tested quickly in frontend.

### DIFF
--- a/lib/requests/RequestItemExport.js
+++ b/lib/requests/RequestItemExport.js
@@ -2,6 +2,7 @@ import { csvFormatRows } from 'd3-dsv';
 
 import ArrayUtils from '@/lib/ArrayUtils';
 import { mapBy } from '@/lib/MapUtils';
+import { formatUsername } from '@/lib/StringFormatters';
 import TimeFormatters from '@/lib/time/TimeFormatters';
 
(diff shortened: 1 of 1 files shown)


candu (commented):
LGTM

PR #752, opened 2021-01-20
Chunked and null-safe fetching of locations in Track Requests

### LINKED ISSUE #750: Chunked location loading for Track Requests
**Description**
Right now, Track Requests loads all study requests up front, as well as all locations for those study requests.  Eventually, this will lead to page load failures due to the limit of 100 locations on `/api/locations/byCentreline`.

In addition, after the most recent conflation update, an issue was noticed in local dev where Track Requests would break if _any_ study request has a location not in the `centreline` conflation target.  Since the centreline can change over time, it's important that we can handle `null` results from `/api/locations/byCentreline` here.

**Acceptance Criteria**
- [x] locations in Track Requests are loaded in chunks, at most 100 at a time;
- [x] `/api/locations/byCentreline` validation schemas are updated to allow for `null` results;
- [x] Track Requests can handle `null` results from `/api/locations/byCentreline`.

**Additional Notes**
This is likely necessary for Version 1.1 - based on the current rate at which users are submitting study requests, we have at most a few weeks before the 100 location limit is reached.

Sending `null` back from `/api/locations/byCentreline` is, of course, not ideal.  It would be better if we could look up the historical centreline feature, but to do that we would need some way of storing / handling historical centreline changes.  @aharpalaniTO or @jwrcoleman may have thoughts here.  (Even with historical centreline support, though, we should be able to handle a `null` result.)

candu:
# Issue Addressed
This PR closes #750 .

# Description
We introduce `ArrayUtils.chunk`, and use it to break the locations list into chunks of no more than `CompositeId.MAX_FEATURES` features.  We also update `/api/locations/byCentreline` to filter out `null` values.

# Tests
Added unit test for `ArrayUtils.chunk`.  Tested quickly in frontend.

### DIFF
--- a/lib/controller/LocationController.js
+++ b/lib/controller/LocationController.js
@@ -157,7 +157,8 @@ LocationController.push({
   handler: async (request) => {
     const { s1 } = request.query;
     const features = CompositeId.decode(s1);
-    return CentrelineDAO.byFeatures(features);
+    const locations = await CentrelineDAO.byFeatures(features);
+    return locations.filter(location => location !== null);
   },
 });
 
--- a/lib/controller/LocationController.js
+++ b/lib/controller/LocationController.js
@@ -148,7 +148,9 @@ LocationController.push({
   options: {
     auth: { mode: 'try' },
     response: {
-      schema: Joi.array().items(CentrelineLocation),
+      schema: Joi.array().items(
+        Joi.object().keys(CentrelineLocation).allow(null),
+      ),
     },
     validate: {
       query: CentrelineSelection,
@@ -157,8 +159,7 @@ LocationController.push({
   handler: async (request) => {
     const { s1 } = request.query;
     const features = CompositeId.decode(s1);
-    const locations = await CentrelineDAO.byFeatures(features);
-    return locations.filter(location => location !== null);
+    return CentrelineDAO.byFeatures(features);
   },
 });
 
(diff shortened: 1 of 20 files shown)


lib/controller/LocationController.js
```diff
@@ -157,7 +157,8 @@ LocationController.push({
   handler: async (request) => {
     const { s1 } = request.query;
     const features = CompositeId.decode(s1);
-    return CentrelineDAO.byFeatures(features);
+    const locations = await CentrelineDAO.byFeatures(features);
+    return locations.filter(location => location !== null);
```
candu, line 161: This change will affect any frontend callsites of `getLocationsByCentreline` _or_ its single-location counterpart `getLocationByCentreline`.  It also points out that `CentrelineDAO.byFeatures` can return `null` values, which we might need to check for in several places (to prevent future bugs with similar causes).

Also: is it better to filter these out, or allow `null` return values?  If the former, should we do that here, or in `CentrelineDAO.byFeatures`?  This may need a bit more thought before being merged, and possibly also some additional test cases in our DB and REST API tests to ensure whatever change we make here doesn't break something.

candu (commented):
Need to address big question around where (and if) to perform filtering.

candu (commented):
Looks good, please test all affected features in frontend again

candu:
Tested features in frontend - LGTM.

PR #754, opened 2021-01-20
Fix text entry bug in date picker

### LINKED ISSUE #729: Unable to edit date in date picker by typing
**What Happened**
Two characters into typing my desired date, the current date overrides into the field, and I can't delete the date that was entered. Any numbers I enter are appended to the date instead of letting me enter the date myself.

**What Should Happen**
I can type the date into the entry field ex. for dates where I'd have to scroll through hundreds of months to get to my starting date

**To Reproduce**
Filter collisions by date:
1. Select an intersection
2. Click on ""Filter"" under ""Collisions""
3. Select the checkbox for ""Filter by date?""
4. Enter a date by writing in the box (e.g. last 10 years - want to look for collisions from 2010-12-09 to 2020-12-09)

**Additional Notes**
This...might be a good candidate for a UI test?  (If so, it would be our first.)

candu:
# Issue Addressed
This PR closes #729 .

# Description
It turns out that `19` is a valid ISO datetime string, which is interpreted as "today at 19:00".  This meant that, whenever the user entered the first two digits of a year, it would detect that as a valid date and fill in the date picker with today's date.

Checking this against a full date regex fixes the issue.

# Tests
Tested quickly in frontend.

### DIFF
--- a/web/components/inputs/FcDatePicker.vue
+++ b/web/components/inputs/FcDatePicker.vue
@@ -52,6 +52,8 @@
 import DateTime from '@/lib/time/DateTime';
 import TimeFormatters from '@/lib/time/TimeFormatters';
 
+const REGEX_VALUE_FORMATTED = /[0-9]{4}-[0-9]{2}-[0-9]{2}/;
+
 function fromInternalValue(internalValue) {
   if (internalValue === null) {
     return null;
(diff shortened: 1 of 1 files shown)


candu (commented):
LGTM

Issue #682, opened 2020-10-29
TMC reports failing to load at St Clair and Oriole


candu:
**What Happened**
When attempting to load the TMC Summary Report at [St Clair and Oriole](https://move.intra.dev-toronto.ca/view/location/s1:AahsmB/POINTS/reports/TMC), the report does not load, and the loading spinner spins indefinitely.

**What Should Happen**
The report should load successfully.

**To Reproduce**
Steps to reproduce the bug:
1. Go to [TMC reports at St Clair and Oriole](https://move.intra.dev-toronto.ca/view/location/s1:AahsmB/POINTS/reports/TMC).

At this point, the report does not load, and devtools report HTTP 500 from the `reporter` backend.

**Additional Notes**
All other TMC report types seem to work fine here; only the summary report fails.

Stacktrace from `reporter` in DEV0:

```
Debug: internal, implementation, error
    TypeError: Cannot destructure property 'data' of 'undefined' as it is undefined.
    at /home/ec2-user/flashcrow/lib/reports/ReportCountSummaryTurningMovement.js:22:54
    at Array.map (<anonymous>)
    at Function.sumIndices (/home/ec2-user/flashcrow/lib/reports/ReportCountSummaryTurningMovement.js:22:47)
    at Function.sumSection (/home/ec2-user/flashcrow/lib/reports/ReportCountSummaryTurningMovement.js:63:51)
    at Function.transformCountData (/home/ec2-user/flashcrow/lib/reports/ReportCountSummaryTurningMovement.js:143:50)
    at ReportCountSummaryTurningMovement.transformData (/home/ec2-user/flashcrow/lib/reports/ReportCountSummaryTurningMovement.js:187:53)
    at ReportCountSummaryTurningMovement.generate (/home/ec2-user/flashcrow/lib/reports/ReportBase.js:126:34)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async handler (/home/ec2-user/flashcrow/lib/controller/ReportController.js:120:26)
```

candu:
Most TMCs in FLOW have 32 data rows: one for every 15-minute block of an 8-hour count.  This location only has 31 data rows:

![St Clair and Oriole only has 31 data rows, not 32](https://user-images.githubusercontent.com/151605/98726688-d7d9a980-2364-11eb-8f14-3fa0696ec132.png)

[...]

PR #756, opened 2021-01-21
Signal warrant: use average for combination, copy update

candu:
# Issue Addressed
This PR makes progress on #682 .

# Description
Turns out that "Traffic Signal Control" should be "Traffic Control Signal", and that divisional practice for the combination warrant (Section 4) is to use the average compliance over the study period - not the minimum compliance of any hour.

# Tests
Tested quickly in frontend, and updated reporting unit tests accordingly so that those still pass.


candu (commented):
LGTM

PR #757, opened 2021-01-22
Email notification revamp

### LINKED ISSUE #748: Expand use of email notifications throughout study request lifecycle
**Description**
Right now, we only send an email to users when they create a new study request.  These emails are not sent to Data Collection by default.

This task aims to expand upon that, bringing a robust system of email notification into the whole study request lifecycle.

**Acceptance Criteria**
See email spec on Notion.

candu:
# Issue Addressed
This PR closes #748 .

# Description
We refactor the `EmailBase` hierarchy, introducing `EmailStudyRequestBase` and `EmailStudyRequestBulkBase`, and a slew of specific email subclasses under each.  We add testing for all these subclasses, and expand checks for `Mailer.send` calls in our REST API tests.  We then update our study request creation, commenting, and update REST API endpoints to send emails as appropriate.

# Tests
New unit and REST API tests.  Will test end-to-end directly in AWS dev, as we do not currently have a way to test delivery itself in local development (only that subjects, recipients, parameters, etc. are as expected).

### DIFF
--- a/lib/email/EmailBase.js
+++ b/lib/email/EmailBase.js
@@ -7,7 +7,6 @@ import { NotImplementedError } from '@/lib/error/MoveErrors';
 const PORT_HTTPS_DEFAULT = 443;
 const {
   emailSender,
-  ENV,
   PUBLIC_PATH,
   session: { domain = 'localhost' },
   webPort = PORT_HTTPS_DEFAULT,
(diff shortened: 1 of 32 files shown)


lib/email/EmailBase.js
```diff
@@ -83,6 +86,27 @@ class EmailBase {
     return emailSender;
   }
 
+  /**
+   * Returns the study request admin email address for this environment.  We
+   * redirect this to an internal MOVE list in non-production environments, to avoid
+   * spamming the Data Collection team with test requests!
+   */
+  static getRecipientStudyRequestAdmin() {
+    if (ENV === 'production') {
```
candu, line 95: this doesn't do what you think it should!  `ENV` here is the _application_ environment, not the _infrastructure_ environment - and the `production` _application_ environment refers to anything that's not local or CI / CD testing!

this should instead check `domain`.

candu (commented):
one important fix to make, otherwise good

candu:
LGTM

Issue #791, opened 2021-02-10
a11y: additional cues to screen readers using sr-only


candu:
**What Should Happen**
Where buttons and other interactive elements depend on context from the surrounding section to make their purpose clear, we should ensure screen reader users have access to textual versions of that same context.  One way to do this is to use the `.sr-only` class to define text that is not displayed, but that will be read out by screen readers.

**JIRA Tasks**
- [x] MOVE-187
- [x] MOVE-182
- [x] MOVE-168
- [x] MOVE-153
- [x] MOVE-151
- [x] MOVE-144
- [x] MOVE-141
- [x] MOVE-134
- [x] MOVE-126
- [x] MOVE-116
- [x] MOVE-72
- [x] MOVE-51
- [x] MOVE-50
- [x] MOVE-49
- [x] MOVE-25
- [x] MOVE-17
- [x] MOVE-12

Issue #783, opened 2021-02-09
Remember Track Requests filters across page transitions


candu:
**Description**
From user report:

> 1. Navigate to the "Track Requests" page from the sidebar
> 2. On getting there, the "User" filter is automatically applied
> 3. Click on the "User" filter chip to deselect that filter, and view all requests
> 4. Click into a request detail
> 5. Click on the "< Requests" button to return to the list of all requests
>
> Expected: I can see the full list of requests, without the "User" filter applied
> Actual: The "User" filter is applied again, and the list is empty. I expect the filters to be remembered.

**Acceptance Criteria**
- [x] move filters state here into `vuex` store;
- [x] when the user first enters Track Requests, set default filters according to user permission;
- [x] when user updates filters, update them in the `vuex` store;
- [x] when the user returns to Track Requests, use the filters from the `vuex` store.

**Additional Notes**
We _could_ persist these filters somewhere like `localStorage` / `sessionStorage`, but that might lead to complications if we ever change the filters here.  For now, we can just handle the case where a user revisits Track Request within the same page load (i.e. without refreshing the browser in between).

PR #796, opened 2021-02-17
Toasts and screen-reader-only text

### LINKED ISSUE #769: Infinite redirect loop from ADFS
**What Happened**
From user report (thanks @mkewins !):

> "I believe I'm getting caught in a redirect loop when my password is incorrect. I might be using the wrong password for our QA testing account since it was updated recently.

> Redirected back to MOVE and immediately back to the login page, with no error/status message.

**What Should Happen**
> Redirected back to MOVE with an error message saying my password is incorrect, OR the login page says password is incorrect.

**To Reproduce**
On AWS dev:
1. Click on ""Manage Downloads"" in the side nav bar
2. Try and login as ""child\esavage"" (our test account)

candu:
# Issue Addressed
This PR makes significant progress on #791 and #783 , and closes #769 .

# Description
We use `FcToastInfo` in several new places to add notifications of various state / mode changes throughout the application.  We also add `span.sr-only` elements to provide additional context to screen reader users.

In the process, we make progress on saving Track Requests filters across page transitions, and address an infinite-redirect bug when the ADFS login flow fails or is abandoned.

# Tests
Tested quickly in frontend.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -183,6 +183,11 @@ HttpStatus.init({
   },
 });
 
+const KeyCode = {
+  ENTER: 0x0a,
+  SPACE: 0x20,
+};
+
 class LegendMode extends Enum {}
 LegendMode.init(['FOCUS_LOCATIONS', 'NORMAL']);
 
@@ -1031,6 +1036,7 @@ const Constants = {
   FEATURE_CODES_INTERSECTION,
   FEATURE_CODES_SEGMENT,
   HttpStatus,
+  KeyCode,
   LegendMode,
   LocationMode,
   LocationSearchType,
@@ -1073,6 +1079,7 @@ export {
   FEATURE_CODES_INTERSECTION,
   FEATURE_CODES_SEGMENT,
   HttpStatus,
+  KeyCode,
   LegendMode,
   LocationMode,
   LocationSearchType,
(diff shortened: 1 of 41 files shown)


lib/Constants.js
```diff
@@ -183,6 +183,11 @@ HttpStatus.init({
   },
 });
 
+const KeyCode = {
```
candu, line 186: can we document the purpose of this?  in general `Constants` has very little documentation, despite being the most widely used / referred-to module.

lib/controller/AuthController.js
```diff
@@ -96,10 +96,10 @@ AuthController.push({
     try {
       const user = await client.callback(request);
       await login(request, user);
-      return h.redirect(config.PUBLIC_PATH);
+      return h.redirect(`${config.PUBLIC_PATH}?login=1`);
```
candu, line 99: please document this as well.  the purpose of this flag isn't immediately clear, especially since it works in combination with frontend logic in `LoginState`.

lib/requests/RequestDataTableColumns.js
```diff
@@ -4,7 +4,8 @@
 const RequestDataTableColumns = [
```
candu, line 4: maybe document this as well?  in particular:

- what does `data-table-expand` do?
- why do we use `text: ''` for some columns, and what are a11y requirements around that?

candu (commented):
couple of places that need more documentation, otherwise good

PR #797, opened 2021-02-18
aria-live div on DOM load

### LINKED ISSUE #791: a11y: additional cues to screen readers using sr-only

candu:
# Issue Addressed
This PR closes #791 .

# Description
We add a new `<div aria-live="polite" role="status">` to our `index.html` page, and update it whenever `FcToast` text is shown / updated.

# Tests
Quickly tested in frontend.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -12,6 +12,7 @@
     <noscript>
       <strong>MOVE requires JavaScript support.  Please enable JavaScript in your browser settings to continue.</strong>
     </noscript>
+    <div aria-live="polite" class="sr-only" id="aria_notification" role="status"></div>
     <div id="app"></div>
     <!-- built files will be auto injected -->
   </body>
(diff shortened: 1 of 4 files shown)


candu (commented):
LGTM

PR #807, opened 2021-02-19
Hide volume data and layer

### LINKED ISSUE #793: Hide AADT numbers and volume estimates
**Description**
From [Notion (internal only)](https://www.notion.so/bditto/Hide-AADT-numbers-and-volume-estimates-169e8793872545769549ef83c270c116):

> It's unclear in MOVE's interface, and people are using this number for analytical purposes (which they shouldn't be right now). To avoid confusion and incorrect use of these estimates, we should hide the AADT numbers and volume layer until Charles / Big Data can revisit how AADT estimates are calculated.

**Acceptance Criteria**
- [x] Remove (or hide) the "Volume" layer from the legend
- [x] Remove (or hide) the teal lines from the map
- [x] Remove (or hide) `AADT (est. 2018)` header and counts from the tooltips
- [x] leave backend pieces intact

candu:
# Issue Addressed
This PR closes #793 .

# Description
We remove the Volume layer from the legend, hide it on the map (by setting `volume: false` in layer settings), and remove AADT estimates from midblock popups.  However, we also leave the backend pieces in place, so that we can quickly roll this back out once AADT estimates are more accurate.

# Tests
Tested quickly in frontend.

### DIFF
--- a/web/components/FcPaneMapPopup.vue
+++ b/web/components/FcPaneMapPopup.vue
@@ -147,15 +147,7 @@ function getCentrelineDescription(feature, { location }) {
   const locationStr = getLocationDescription(location);
   description.push(locationStr);
 
-  const { centrelineType } = feature.properties;
-  if (centrelineType === CentrelineType.SEGMENT) {
-    let { aadt = null } = feature.properties;
-    if (aadt !== null) {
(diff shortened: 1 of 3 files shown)


candu (commented):
LGTM

Issue #792, opened 2021-02-17
Peak hour factor reports


candu:
**Description**
As described [on Notion (internal only)](https://www.notion.so/bditto/Peak-Hour-Factor-PHF-Reports-036b271abd244b10933faff65aaa1143), MOVE is currently missing the Peak Hour Factor report available in FLOW.

This task adds that report type for all turning movement counts.

**Acceptance Criteria**
- [x] add peak hour factor report module to `reporter`;
- [x] show that report type in the interface;
- [x] add unit tests for `ReportTimeUtils`;
- [x] add a side-by-side test for this report type.

candu:
There isn't time at this point to add a side-by-side test.  I'll merge the work we have, and move this to the bottom of the priority list.

PR #808, opened 2021-02-22
Peak hour factor reports: implementation

candu:
# Issue Addressed
This PR makes significant progress on #792 .

# Description
We introduce `ReportPeakHourFactor`, a new `ReportBaseFlow` subclass for generating peak hour factor reports.  We also create `ReportTimeUtils`, which contains basic utilities for finding AM, PM, peak periods, and consecutive hours from count data.

# Tests
Tested quickly in frontend.  (The reason this doesn't close out the related issue is because I haven't yet written a side-by-side test for this.)


lib/reports/ReportPeakHourFactor.js
```diff
@@ -0,0 +1,323 @@
...
+import {
+  indexRangeAm,
+  indexRangeMax,
+  indexRangePeakHour,
+  indexRangePm,
+  indexRangeSum,
+} from '@/lib/reports/time/ReportTimeUtils';
+import TimeFormatters from '@/lib/time/TimeFormatters';
+
+const dirs = ['N', 'E', 'S', 'W'];
```
candu, line 16: we should ideally use `CardinalDirection.enumValues.map(({ short }) => short)` and `TurningMovement.enumValues.map(({ short }) => short)` here.  (same for a few of the other modules that hardcode these - it's easier if we can unify this across all report types!)

lib/reports/ReportPeakHourFactor.js
```diff
@@ -0,0 +1,323 @@
...
+    return peakData;
+  }
+
+  static getPeakHourFactor(countData, indexRangePeak, fnVolume) {
+    const volumeTotal = indexRangeSum(countData, indexRangePeak, fnVolume);
+    const volumeMax = indexRangeMax(countData, indexRangePeak, fnVolume);
+    if (volumeMax === 0) {
+      return null;
+    }
+    return volumeTotal / (4 * volumeMax);
```
candu, line 50: this still assumes that `countData` contains 15-minute buckets, but we can revisit that assumption more fully later.  (please document it for now, though, so that we have an easier time changing this later if need be.)

lib/reports/time/ReportTimeUtils.js
```diff
@@ -0,0 +1,191 @@
+/**
```
candu, line 1: we should at least add some unit tests for these, even if we're not going to write a side-by-side test yet.  (maybe we can add that to the issue, and write them up when we circle back to it?)

candu (commented):
mostly good, a couple of small comments

Issue #803, opened 2021-02-19
Small differences between email notifications implementation and spec


candu:
**What Happened**
There are a few small differences between our email notifications implementation and the spec.

**What Should Happen**
After discussion with @mkewins , we have a few specific changes:

- [x] for new comments, add the commenter's username to the email;
- [x] for bulk requests:
  - [x] do not link to the multi-location selection;
  - [x] ensure that we link to the bulk request somewhere;
  - [x] include the individual study request locations, as well as a link to the individual requests;
  - [x] remove bullet points;
- [x] for bulk request completed / cancelled:
  - [x] send individual notifications for cancelled requests;
  - [x] otherwise, do not send individual request emails (e.g. for completed)
  - [x] when the whole bulk request is completed (or in some end state), send bulk notification.

**To Reproduce**
See testing notes for steps here.

**Additional Notes**
We'll need a way to determine when the bulk request is finished.  (It might be a good idea to update how `bulkStatus` works, since this is also used to show bulk request status in the status bar!  We can also add a `terminal` flag to `StudyRequestStatus` codes.)

In this case, we can think of cancelled and completed as `terminal` - we now know that when data is rejected, the request is often resubmitted to the contractor for collection.  This means that a rejected request will eventually end up in either completed or cancelled, i.e. rejected is not `terminal`.  In other words: we can consider a bulk request `terminal` when all of its requests have a `terminal` status.

candu:
Tested quickly in AWS dev, marking closed.

PR #810, opened 2021-02-22
Small tweaks to email notifications

candu:
# Issue Addressed
This PR makes significant progress on #803 .

# Description
We update email notification copy to fix a bug where per-request locations weren't showing up for bulk request emails, and add commenter details.  In the process, we refactor the common list of requests in a bulk request out into a reusable template string.

We also update `MailUtils` to reduce the volume of emails sent around bulk requests.

Finally, we add more testing of `MailUtils` and various `EmailBase` subclasses.

# Tests
Ran additional unit tests, plus the usual `ci:jest-coverage` check.  Will have to test end-to-end delivery in AWS dev.


candu (commented):
LGTM

PR #811, opened 2021-02-23
One-time userOnly initialization for Track Requests

### LINKED ISSUE #783: Remember Track Requests filters across page transitions

candu:
# Issue Addressed
This PR closes #783 .

# Description
We initialize `userOnly` once in `FcRequestsTrack`.  We also make one small tweak to `RequestActions`: to avoid confusion, we disable the current status in the status menu (except for `ASSIGNED`, so that staff can still reassign things).

# Tests
Tested quickly in frontend, plus CI checks to ensure `RequestActions` changes don't break REST API.

### DIFF
--- a/lib/requests/RequestActions.js
+++ b/lib/requests/RequestActions.js
@@ -44,7 +44,11 @@ class RequestActions {
   }
 
   static canCancel(user, studyRequest) {
-    return RequestActions.canEdit(user, studyRequest);
+    if (!RequestActions.canEdit(user, studyRequest)) {
+      return false;
+    }
+    const { status } = studyRequest;
+    return status !== StudyRequestStatus.CANCELLED;
   }
 
   static canEdit(user, studyRequest) {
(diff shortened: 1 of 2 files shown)


candu (commented):
LGTM

PR #812, opened 2021-02-23
Fix non-ASCII characters in locations, camel-case header names

### LINKED ISSUE #802: Encoding and column name issues with exported list of requests
**What Happened**
From QA notes (thanks @mkewins !):

> the dash still encoded funny
> - excel: `Wright Ave: Ln 2 W Sorauren N Fern â€“ Roncesvalles Ave`
> - in move: `Wright Ave: Ln 2 W Sorauren N Fern – Roncesvalles Ave`

> I also noticed the headers are formatted well, up until the last ~5 or so, which all have camel case named (pulled directly from variables?)

**What Should Happen**
CSV exports of requests should contain only ASCII characters, to prevent encoding issues with Excel.  Headers in this CSV export should be human-readable, and not use camel-case.

**To Reproduce**
See testing notes for details.  (The dash should be present for most midblocks, and the headers should be the same for any set of exported requests.)

**Additional Notes**
I recall someone (@aharpalaniTO ?  @jwrcoleman ?) mentioning encoding issues with Excel being a common problem.  (IIRC it was that it doesn't use UTF-8 as a default, but instead uses some old Windows code page that's similar to ASCII.

candu:
# Issue Addressed
This PR closes #802 .

# Description
We make some quick fixes to `RequestItemExport` to address issues noted from internal QA testing.

# Tests
Tested on set of requests covering both single and bulk, and both intersection and midblock locations.

### DIFF
--- a/lib/requests/RequestItemExport.js
+++ b/lib/requests/RequestItemExport.js
@@ -30,14 +30,14 @@ const CSV_COLUMNS = [
    *
    * We also include several columns that are not included in the above list.
    */
-  'id',
-  'assignedTo',
-  'status',
-  'urgent',
-  'urgentReason',
-  'lng',
-  'lat',
-  'daysOfWeek',
+  'MOVE ID',
+  'Assigned To',
+  'Status',
+  'Urgent',
+  'Urgent Reason',
+  'Longitude',
+  'Latitude',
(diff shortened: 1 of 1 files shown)


candu (commented):
LGTM

Issue #798, opened 2021-02-18
Address feedback on initial a11y changes


candu:
**Description**
After the big sweep of a11y tasks over the last couple of weeks, @shinechaudhuri and @mkewins provided some [great feedback (internal only)](https://www.notion.so/bditto/Feedback-on-initial-a11y-changes-1329ecad365d485790202d8296eac9b0) on these updates from a design / product perspective.

This task addresses that feedback.

**Acceptance Criteria**
- [x] remove gray "total count" boxes for Collisions and Studies on View Data
- [x] reduce size of `<h2>` headings to save space: View Data, Track Requests, View Request, Request Study
- [x] change "Details" to "Request Details" in View Request
- [x] add left margin spacing between action buttons and title in View Request header
- [ ] revamp top navigation:
  - [ ] move side navigational elements into top bar
  - [ ] add Back button to top bar
  - [x] use `border` instead of `box-shadow`
  - [ ] get rid of title?
- [ ] from [MOVE-130](https://toronto.atlassian.net/browse/MOVE-130): rethink headers on non-map pages

**Additional Notes**

candu:
Marking closed for now - we can open more specific tasks on the navigational pieces as needed, but if we're planning to keep both top and side anyways we can likely make do with what we have for next release.

PR #813, opened 2021-02-23
Address remaining axe issues, define progress / tooltip wrappers

candu:
# Issue Addressed
This PR makes further progress on #798 .

# Description
We run another axe sweep over major parts of MOVE, fixing things along the way.  In the process, we define a few new wrapper components for accessible progress indicators and tooltips.

# Tests
Tested quickly in frontend, including a sweep with the axe Chrome extension.

### DIFF
--- a/web/components/inputs/FcDatePicker.vue
+++ b/web/components/inputs/FcDatePicker.vue
@@ -8,8 +8,8 @@
       min-width="290px"
       offset-y
       transition="scale-transition">
-      <template v-slot:activator="{ attrs, on: onMenu }">
-        <div v-bind="attrs">
+      <template v-slot:activator="{ attrs: attrsMenu, on: onMenu }">
+        <div v-bind="attrsMenu">
           <v-text-field
             v-model="valueFormatted"
             append-icon="mdi-calendar"
--- a/web/components/inputs/FcDatePicker.vue
+++ b/web/components/inputs/FcDatePicker.vue
@@ -9,33 +9,35 @@
       offset-y
       transition="scale-transition">
       <template v-slot:activator="{ attrs, on: onMenu }">
-        <v-text-field
-          v-model="valueFormatted"
-          append-icon="mdi-calendar"
-          offset-y
-          v-bind="{ ...attrs, ...$attrs }"
-          @blur="resetValueFormatted"
-          @input="updateValueFormatted"
-          @click:append="showMenu = !showMenu"
-          v-on="onMenu">
-          <template v-slot:append>
-            <v-tooltip right>
-              <template v-slot:activator="{ on: onTooltip }">
-                <v-icon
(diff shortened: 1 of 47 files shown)


web/components/inputs/FcDatePicker.vue
```diff
@@ -9,33 +9,35 @@
       offset-y
       transition="scale-transition">
       <template v-slot:activator="{ attrs, on: onMenu }">
```
candu, line 11: can we do a similar `attrs: attrsMenu` thing here?  that would help make it clear that these are a11y attributes specifically related to the `<v-menu>` activator slot

web/components/inputs/FcDatePicker.vue
```diff
@@ -9,33 +9,35 @@
...
+            v-bind="$attrs"
+            @blur="resetValueFormatted"
+            @input="updateValueFormatted"
+            @click:append="showMenu = !showMenu"
+            v-on="onMenu">
+            <template v-slot:append>
+              <FcTooltip right>
+                <template v-slot:activator="{ on: onTooltip }">
+                  <v-icon
+                    aria-label="Select date using calendar"
```
candu, line 26: we might have to add `:aria-hidden="false"` here

web/components/inputs/FcInputLocationSearch.vue
```diff
@@ -9,57 +9,56 @@
       :offset-y="true"
       :open-on-click="false">
       <template v-slot:activator="{ attrs, on: onMenu }">
```
candu, line 11: same here: use `attrs: attrsMenu`.

we should also check for this same pattern elsewhere - do we have other inputs that need to be wrapped in a `<div>`, to avoid incorrect use of ARIA roles?

web/components/nav/FcDashboardNavUser.vue
```diff
@@ -49,7 +49,7 @@
         </v-list-item>
       </v-list>
     </v-menu>
-    <v-tooltip
+    <FcTooltip
```
candu, line 52: we can pull these attributes up onto a single line here

web/views/FcLayoutDrawerMap.vue
```diff
@@ -17,10 +17,9 @@
           left>{{iconDrawerToggle}}</v-icon>
         {{labelDrawerToggle}}
       </FcButton>
-      <v-tooltip
+      <FcTooltip
```
candu, line 20: same here, can pull up onto single line for readability

candu (commented):
mostly good, a few small changes

web/components/inputs/FcDatePicker.vue
candu: nvm, Vuetify adds useful ARIA roles here (probably because this is used as an activator target?)

PR #816, opened 2021-02-24
"Other" types for study requests

### LINKED ISSUE #795: Study requests: additional study types
**Description**
Several users have tried to request study types that are not currently included in our `StudyType` enum.  We've also made incorrect assumptions about where some study types are performed - for instance, Ped Delays.

This task adds a couple of frequently-requested study types to our list, and also adds an "Other" option for cases where an unusual or new study type needs to be performed.

**Acceptance Criteria**
- [x] allow users to submit requests for Ped Delays at intersections;
  - [x] Ped Delays: hours or duration?
- [x] add "Other" option for study type;
- [x] allow users to create "Other" study requests at both intersections and midblocks;
- [x] when users select "Other" as the study type:
  - [x] show a text field that allows them to enter the study type, and make that field required;
  - [x] hide both hours and duration, and force the user to enter additional details;
- [x] add a `studyTypeOther` field to database records, and default to `NULL` for existing study requests;
- [x] ensure that `studyTypeOther` is in validations, and that test cases include this field.

**Additional Notes**
This strikes a deliberate balance between offering admin management of study types and our current "only these study types can be used" approach.  Under this new approach, study types can be "promoted" over time through subsequent releases:

- first, a new study type will be entered under "Other";
- if many people are requesting this study type, we can add an explicit option for it, and go back and mark those requests with that new option;
- eventually, MOVE will also incorporate data, reports, etc. from that type of study.

candu:
# Issue Addressed
This PR closes #795 .

# Description
We introduce `StudyType.OTHER_AUTOMATIC` and `StudyType.OTHER_MANUAL`, two new "other" study types designed to allow users to request studies that aren't covered by the list of study types offered in MOVE.

To support this, we also introduce the new `studyTypeOther` field for study requests, where users can enter a brief description of the type of study to be performed.

# Tests
Updated a bunch of study-request-related unit, DB, and REST API tests to test this new field and all relevant flows / validations.  Tested end-to-end briefly in browser, by creating a study request of one of these new "other" types and ensuring it can be viewed / edited successfully.


tests/jest/unit/model/StudyRequestBulk.spec.js
```diff
@@ -21,6 +21,7 @@ test('StudyRequest', () => {
     reasonOther: null,
     ccEmails: [],
     studyType: StudyType.TMC,
+    studyTypeOther: null,
```
candu, line 24: let's use `generateStudyRequest()`, `generateStudyRequestBulk()` here as well (and in `StudyRequest.spec.js`)

candu (commented):
mostly good, small comment to use study request generators in model tests

Issue #778, opened 2021-02-04
Different collision numbers from CRASH


mkewins:
**What Happened**
Karen reported a discrepancy in collisions - total and verified - between CRASH and MOVE along Sheppard from Beecroft (`px:1169`) to Leslie (`px:744`).

**What Should Happen**
I'm unsure whether the numbers should match exactly, but we should get to the bottom of why they're different.

**To Reproduce**
Steps to reproduce:
1. In multi-location mode, select a corridor from `Sheppard / Beecroft` to `Sheppard / Leslie` (inclusive)
2. Save the selection and go to the View Data page
3. Filter collisions by date: 1/1/2015 to 12/31/2019
4. Observe 2665 total collisions and 608 verified

For the same selection and date filter, Karen reported 495 verified and 2630 total. (I haven't been able to confirm this because I'm struggling with CRASH.)

A very specific example of this difference can be seen at `Sheppard Ave E: Doris Ave – Bonnington Pl` for the same date range:
* MOVE reports 3 verified / 5 total
* CRASH reports 2 verified / 5 total

The collision ID numbers are the same for the five collisions. The only difference is the collision that occurred in December 2015 was marked as verified in MOVE, but not in CRASH.

**Screenshots**
n/a

**Additional Notes**
In CRASH, I noticed a small line of text that indicates CRASH uses a 20m radius around intersections, where I believe we're using a 30m radius in MOVE.

candu:
Found additional evidence suggesting that CRASH uses a 20m buffer - the CRASH source code refers to a setting `buffer_size_m`, which is loaded from an MS Access database (!) in the application folder called `Local.mdb`.  I couldn't open the file in MS Access itself, as it was created using an ancient version that's no longer supported; however, a quick search of this file in a text editor suggests `buffer_size_m` is set to 20m.

I'll update the pipelines accordingly.

candu:
After changing the buffer radius to 20m, the total number of collisions is a bit closer in MOVE (2653 now), but the number of validated collisions remains high as compared to MOVE.

Found what looks like the source of the discrepancy between 3 / 5 verified (in MOVE) and 2 / 5 verified (in CRASH), by looking at the database records directly.  (I tried locating the logic in CRASH source code, but couldn't manage to track that down.)

In `collisions.events` (MOVE) / `"TRAFFIC"."ACC"` (CRASH), there are 3 possible values for `changed`:

- `NULL`: this record has never been verified;
[...]

Issue #775, opened 2021-02-03
Missing signalized intersection - Steeles between Woodbine and Victoria Park


mkewins:
**What Happened**
MOVE shows one flush midblock where there is a signalized intersection (you can see traffic lights in the "aerial" image in MOVE) between Woodbine and Victoria Park, on Steeles.

**What Should Happen**
This segment should be displayed as two midblocks, and one intersection.

**To Reproduce**
Steps to reproduce the bug:
1. Go to Steeles/Victoria Park (px: 1142)
2. Scroll to the midblock immediately left/west

**Screenshots**
![Screenshot 2021-02-03 102240](https://user-images.githubusercontent.com/3179624/106768311-cc1b1d80-6609-11eb-9824-7e53c6cedbed.png)

**Additional Notes**
n/a

mkewins:
Shawn Dillon reported another missing intersection at Lansdowne / Lansdowne.

The intersection is located south of `Lansdowne Ave / Dundas St W` and north of `Lansdowne Ave / Shirley St`.

![Shawn Dillon - Missing Intersection](https://user-images.githubusercontent.com/3179624/106818663-98aab400-6646-11eb-8bdf-001bef7ce7c4.png)

candu:
Looking deeper into this: in both cases, the issue is that the intersection isn't in the centreline, so fixing this would mean either a) maintaining our own patch to the centreline or b) getting it fixed upstream somehow.

For the Steeles and Victoria Park case: the road south off of Steeles looks to be a private entrance to some sort of technology / office park - private driveways, roads, etc. aren't in the centreline; only City-maintained roads are.

For the Lansdowne / Lansdowne intersection: not sure why this isn't in the centreline, but it's not; maybe the intersection just north of the circled area is intended to capture this leg as well?

I'm not sure what our approach _should_ be here - in general, it's unlikely we'll _ever_ achieve 100% accuracy, no matter what data we use for the road network, and using the Centreline as-is gives us a lot of value.  This feels like a thing where we may just have to communicate this to our users.

Issue #755, opened 2021-01-21
Corridor routing issues near large parks (Mt. Pleasant, High Park)


candu:
**What Happened**
From a user-submitted bug report:

> There is some link on Yonge Street that is not connected

This causes corridor routing near Mt. Pleasant on Yonge to fail.  A similar issue was observed when trying to reproduce near High Park on Bloor.

**What Should Happen**
These corridors should be routable.

**To Reproduce**
Steps to reproduce the bug:
1. Go to the map, search for "Yonge and St. Clair", and select the intersection
2. Click "Add Location", add "Yonge and Belsize"
3. Check "include intersections and midblocks between locations"
4. See error (red toast at bottom)

**Screenshots**
![Corridor routing error on Yonge near Mt. Pleasant](https://user-images.githubusercontent.com/151605/105370064-2d39fe80-5bd1-11eb-8fb9-14558820d485.png)

![Corridor mis-routed on Bloor near High Park](https://user-images.githubusercontent.com/151605/105370156-48a50980-5bd1-11eb-95b3-4f66cb6c1ed6.png)

**Additional Notes**
For the Bloor near High Park case, these are midblock IDs 30119699, 30119700.  A closer look at centreline geometry in the database reveals a strange quirk:

```sql
flashcrow=> select fnode, tnode from centreline.midblocks where "centrelineId" = 30119699;
  fnode   |  tnode
----------+----------
 13466332 | 30119698
(1 row)

flashcrow=> select fnode, tnode from centreline.midblocks where "centrelineId" = 30119700;
  fnode   |  tnode
----------+----------
 30119698 | 13466395
(1 row)

flashcrow=> select * from centreline.intersections where "centrelineId" = 30119698;
 centrelineId | centrelineType | classification | description | featureCode | geom | lat | lng
--------------+----------------+----------------+-------------+-------------+------+-----+-----
(0 rows)

flashcrow=> select * from centreline.midblocks where "centrelineId" = 30119698;
 aadt | centrelineId | centrelineType | featureCode | fnode | fromIntersectionName | geom | lat | lng | midblockName | roadId | tnode | toIntersectionName
------+--------------+----------------+-------------+-------+----------------------+------+-----+-----+--------------+--------+-------+--------------------
(0 rows)

flashcrow=> select * from gis.centreline_intersection where int_id = 30119698;
[...]

PR #822, opened 2021-03-04
Test updates: TCL v2, validated collision fix 

candu:
# Issue Addressed
This PR makes progress on #775 and #778 .

# Description
We update tests to match our test dataset after updating to TCL v2 and changing the collision buffer radius from 30m to 20m.  We also change the logic for determining when a collision is validated from `changed !== null` to `changed === -1`, to match the (apparent) behaviour of CRASH here.

# Tests
Updated to latest `flashcrow-dev-data.sql` in both local development and our `gitlab-runner` environment on AWS dev.  Ran `npm run ci:jest-coverage` to ensure all tests pass locally.  Used GitLab CI pipelines to ensure all tests pass there.

### DIFF
--- a/tests/jest/db/RoutingDAO.spec.js
+++ b/tests/jest/db/RoutingDAO.spec.js
@@ -161,7 +161,7 @@ test('RoutingDAO.routeFeatures', async () => {
   featureTo = { centrelineId: 444912, centrelineType: CentrelineType.SEGMENT };
   result = await RoutingDAO.routeFeatures(featureFrom, featureTo);
   expect(result).toEqual({
-    next: { centrelineId: 13454835, centrelineType: CentrelineType.INTERSECTION },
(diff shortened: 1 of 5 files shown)


tests/jest/db/RoutingDAO.spec.js
```diff
@@ -259,6 +261,8 @@ test('RoutingDAO.routeCorridor', async () => {
     { centrelineId: 445100, centrelineType: CentrelineType.SEGMENT },
     { centrelineId: 13455130, centrelineType: CentrelineType.INTERSECTION },
     features[3],
+    { centrelineId: 13455130, centrelineType: CentrelineType.INTERSECTION },
```
candu, line 264: these changes feel a bit weird - why is it doubling back here? - but we can figure it out as part of our routing updates for #755 .

candu (commented):
LGTM

Issue #825, opened 2021-03-05
Usability fixes for Version 1.2 


candu:
**Description**
This is a tracking task for multiple smaller tasks identified through QA and usability testing.

**Acceptance Criteria**
- [ ] [bulk cancel notification issue (internal only)](https://www.notion.so/bditto/Last-bug-for-email-notifications-18d713141baf420c851665446120e54b)
- [x] remove CSV download option from warrant report
- [x] add more padding in View Request: [details (internal only)](https://www.notion.so/bditto/To-discuss-request-page-changes-b768f9e8ec6c40919d005cc971fc863c)
- [x] add full time range to TMC detailed report
- [x] add PX number to map tooltip

**Additional Notes**

PR #830, opened 2021-03-11
Report fixes: time ranges, test refactoring, PHF discrepancies

candu:
# Issue Addressed
This PR makes further progress on #825 and #792 .

# Description
We add a side-by-side test against legacy values for `ReportPeakHourFactor`, and use that to track down a potential cause: we were using multi-modal totals instead of _vehicle_ totals to find the peak hour.  We don't yet have tests for `ReportTimeUtils`, but those are coming.

In the process, we also refactor the report side-by-side tests a bit to move common setup boilerplate for each dataset into `SetupTestData`, similar to what we did for directional datasets previously.

# Tests
Ran side-by-side test, as well as full CI checks.  Will test on [PHF test cases (internal only)](https://www.notion.so/bditto/Peak-Hour-Factor-PHF-Reports-036b271abd244b10933faff65aaa1143#f145e38fbc0d43e2b989a7f321dc8090) after this gets deployed to AWS dev.


candu (commented):
LGTM

PR #857, opened 2021-03-26
Global filters: map integration

### LINKED ISSUE #846: Use global filters to filter on map: all studies, last 10 years collisions
**Description**
Up to this point in our Global Filters implementation, we've introduced the filter box and drawer, but the filters present there still only affect View Data and View Reports.

In this task, we complete an initial implementation of Global Filters by implementing filtering on the map, using all studies and the last 10 years of collision data.

**Acceptance Criteria**
- [x] relevant fields for filtering are added to collision vector tiles:
  - [x] `tippecanoe`-generated tiles in ETL;
  - [x] dynamic tiles from `DynamicTileController`;
- [x] we pass common, collision, and study filters to `GeoStyle.get` from the `mapStyle` computed prop of `FcPaneMap`;
- [x] those filters are used in `GeoStyle` to build filter expressions on relevant layers;
- [x] we detect when the user's filters would include data from further back than 10 years, and show a warning in this case.

**Additional Notes**
While the need to show a warning isn't ideal, at this point we'll at least have an initial working implementation of Global Filters.  We can then separately investigate how to show / filter the entire collision dataset at once.

candu:
# Issue Addressed
This PR closes #846 .

# Description
We update both dynamic and pre-generated collisions vector tiles to include several new fields, then update `GeoStyle` to filter the relevant layers using global filters.  We also update `GeoStyle` tests, adding a random generator for filters so that we can fuzz-test that we're still producing valid styles when filters are applied.

# Tests
New `GeoStyle` unit test for style validity.  Tested in browser end-to-end: vector tiles show up as before, filters affect map when zoomed in (level 1), filters affect map when zoomed out (levels 2, 3).


candu (commented):
LGTM

Issue #858, opened 2021-03-29
Pagination for Track Requests


candu:
**Description**
In production, [Track Requests (internal only)](https://move.intra.prod-toronto.ca/requests/track) still loads all study requests ever made at once.  As of time of writing, that's now over 350 requests, and it takes ~1.5s to load; this is followed by another 2-3s to render the table.  (This was a deliberate choice, under the assumption that the rate of requests was low enough that we'd have time to fix this - but this is a previously known piece of technical debt that's now worth paying off.)

This task aims to fix that by introducing pagination and by performing filter / search / sort operations in the backend, similar to View Data.

**Acceptance Criteria**
- [x] decide on pagination approach;
- [x] decide how pagination should account for bulk vs. non-bulk requests;
- [x] add appropriate pagination controls to the Track Requests interface;
- [x] fix #851 ;
- [x] introduce a new `RequestFilters` model to request filter / search parameters;
- [x] change the `GET /requests/study` endpoint in `StudyRequestController` to accept `RequestFilters` parameters;
- [x] from the Track Requests interface, make requests to `GET /requests/study` with `RequestFilters` parameters.

**Additional Notes**
There currently aren't any bulk study requests in production (!), so interleaving bulk vs. non-bulk requests will have to be tested in dev / QA.

candu:
This has turned out to be a much larger task than expected!  Fortunately, we have a viable path forward for implementation; see [Notion (internal-only)](https://www.notion.so/bditto/Study-Requests-Backend-Pagination-3053fa9624a9458eadbf93111334fa80) for details.

PR #860, opened 2021-04-09
Pagination for Track Requests

candu:
# Issue Addressed
This PR makes significant progress on #858 .

# Description
We move Track Requests filter, search, and sort logic from the frontend to the backend.  In the process, we introduce a new table `study_request_items` as a combined index of bulk and non-bulk requests, and integrate upsert / delete hooks into `StudyRequestDAO` and `StudyRequestBulkDAO` to ensure `study_request_items` is kept up-to-date with changes to `study_requests` and `study_requests_bulk`.

To address location searching, we also switch from the old `MidblockDescription.get` approach to one backed by `location_search.centreline`, which uses a port of `MidblockDescription.get` to pre-populate midblock names.  `CentrelineDAO` is also updated to use this, and we get rid of the JavaScript `MidblockDescription` module.

Finally, we add a new script `generate-study-requests.js` to generate a bunch of test requests at once, making it easier to test this functionality out.

# Tests
Tested filter, search, sort operations in frontend on a set of 1000 test requests generated via `generate-study-requests.js`.  Tested `study_request_items` hook integration by running `generate-study-requests.js`, which in turn runs these hooks.  Still need to add database tests for `StudyRequestItemDAO`.


web/components/FcDataTableRequests.vue
```diff
@@ -100,10 +101,7 @@
           Multiple Locations
           <FcTextNumberTotal
             class="ml-1"
-            :k="hasFilters ? item.studyRequestBulk.studyRequests.length : null"
-            :n="hasFilters
-              ? item.studyRequestsTotal
-              : item.studyRequestBulk.studyRequests.length" />
+            :n="item.studyRequestBulk.studyRequests.length" />
```
candu, line 104: this does mean we're getting rid of the whole "show only matching sub-requests" thing, though tbh I'm not sure how useful that was - will be interesting to get feedback on this change.

web/components/FcDataTableRequests.vue
```diff
@@ -136,15 +134,7 @@
       </span>
     </template>
     <template v-slot:item.ASSIGNED_TO="{ item }">
-      <FcMenuStudyRequestsAssignTo
```
candu, line 139: also this change.  (we can undo it later, but I'd like to really simplify the interface / rendering here - it'll make the whole thing feel much snappier.)

web/components/dialogs/FcDialogRequestFilters.vue
```diff
@@ -78,30 +73,6 @@
             :value="assignee"></v-checkbox>
         </fieldset>
 
-        <FcRadioGroup
```
candu, line 81: same here - I've removed the `createdAt` and `lastEditedAt` filters, mostly because these would have been *hard* to set up with `study_request_items`.

the issue here is that I haven't found an easy way to build queries of the form "is any element in this array within this range?" - we'd have to do something like `unnest()`, but then we'd likely end up doing multiple queries and merging results from them, and that's even more complexity.

web/views/FcRequestsTrack.vue
```diff
@@ -65,18 +65,42 @@
...
             :columns="columns"
             :has-filters="hasFiltersRequest"
+            :height="heightTable"
             :items="items"
+            :items-per-page.sync="itemsPerPage"
             :loading="loading"
-            :sort-by.sync="sortBy"
-            :sort-desc.sync="sortDesc"
+            :page.sync="page"
+            :server-items-length="items.length"
```
candu, line 80: do we need `server-items-length` here?  we're handling all the pagination outside `<FcDataTableRequests>` anyways...

web/views/FcRequestsTrack.vue
```diff
@@ -189,6 +212,15 @@ export default {
       });
       return itemsAll;
     },
+    numPages() {
+      return Math.ceil(this.total / this.itemsPerPage);
+    },
+    pageFrom() {
+      return (this.page - 1) * this.itemsPerPage + 1;
```
candu, line 219: minor bug: if `this.total === 0`, this will still return 1.  we should do `Math.min(0, ...)` here, or even just have an explicit `if (this.total === 0) { return 0; }` check.

candu (commented):
mostly good, but needs more documentation, and a couple of things to clean up

candu (commented):
LGTM

candu:
Note that these are [all failing test coverage checks on `lib/db` (internal only)](https://gitlab.bdit.intra.prod-toronto.ca/move-team/bdit_flashcrow/pipelines) right now.  (That's fine; we just have to make sure we add tests in a follow-up PR.)

Issue #862, opened 2021-04-12
New release: Version 1.2.1


candu:
**Release Checklist**

All CI checks pass:
- [ ] `npm run ci:npm-outdated` (to fix: update `package.json`, `npm install`; if upgrading a package breaks MOVE, add it to `.ncurc.json`)
- [ ] `npm run ci:npm-audit` (to fix: `npm audit fix`; if issues remain, `npx resolve-audit`)
- [ ] `npm run ci:jest-coverage` (to fix: add / improve tests, address test failures!)
- [ ] `npm run frontend:build` (to fix: address build errors)
- [ ] `npm run docs:build` (to fix: address build errors)

Manual browser testing:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Quick check in browser OK, but we'll be QA'ing this more extensively as well.

Before deployment:
- [ ] `CHANGELOG.md` is updated to include details of new release
- [ ] version bumped to `1.2.1` in `package.json` using `npm version patch`
- [ ] above changes reviewed and merged to `master`

Deployment:
- [ ] new version pushed pushed to CodeCommit
- [ ] deployment to AWS succeeds

candu:
Note: test coverage has dropped in `lib/db` below set thresholds, but I'm going to merge / deploy this anyways so we can get early QA testing in.  Going to add an issue in v1.3 release milestone to restore coverage to above thresholds here.

PR #867, opened 2021-04-12
Make a small change to trigger redeploy

candu:
# Issue Addressed
This PR is a follow-up on #862 .

# Description
In this case, we wanted to add `p2` labels on bugs and features by default, so I've just added those to the issue templates.

# Tests
Main purpose here is to try upgrading the application database again, but I'll also check the templates quickly after merge.


candu (commented):
LGTM

Issue #874, opened 2021-04-15
Changes to collision filtering options


candu:
**Description**
We have a first implementation of global filters up on MOVE dev / QA 🎉 

This task makes a few small changes to collision filters, based on follow-up and feedback from @jwrcoleman @mkewins .

**Acceptance Criteria**
Support for the following filters is added:
- [x] injury level: KSI (fatals and minor injuries), other levels in nested structure;
- [x] validation status: validated, not validated, all;
- [x] MVCR status: available, missing, all;
- [x] vehicle type (`vehtype`, IIRC) in nested structure;
- [x] resulting from running red light.

Support is added to:
- [x] `CollisionFilters`;
- [x] `FilterGenerator`;
- [x] `CollisionDAO`;
- [x] `DynamicTileDAO`;
- [x] `GeoStyle`;
- [x] Airflow collision normalization pipelines;
- [x] Airflow collision vector tile pipelines.

Additional tasks:
- [x] order of filters should match that in [Data Filters (internal-only)](https://www.notion.so/bditto/Data-Filters-02f2c742748943b2b666de8fcb34640e);
- [x] rename "Property Damage" to "Damage to City Property";
- [x] rename "Manoeuver" to "Manoeuvre";
- [x] move "Damage to City Property" out of "Emphasis Areas", and rename those areas to "Vision Zero Emphasis Areas".

**Additional Notes**
More research is needed to figure out how to identify collisions that resulted from running a red light.

See [`bdit-data_sources`](https://github.com/CityofToronto/bdit_data-sources/pull/349#issuecomment-803128223) for a discussion of how to distinguish TPS vs. CRC collisions, in case we want to do that at some point.

candu:
@mkewins pointed out that we've already documented red light collisions [here (internal-only)](https://www.notion.so/bditto/CRASH-Filtering-and-Reports-468066f330764a9584ba979e0265da96#56a83615bd9140d1887eb5bfa9261a63).

In View Data, we use `e.changed = -1` to identify validated collisions.  (This used to be `e.changed IS NOT NULL`, but we later learned that a value of `0` indicates that only partial validation has occurred, and that CRASH only counts _fully_ validated records.)

candu:
For MVCR presence / absence: there's `mvaimg`, which _also_ appears to have three possible values of `NULL`, `0`, `-1`.  Maybe we should use similar `e.mvaimg = -1` logic?  This might need to be double-checked against CRASH.

[...]

PR #876, opened 2021-04-16
Collision filters: web frontend / backend updates

candu:
# Issue Addressed
This PR makes significant progress on #874 .

# Description
We refactor collision, common, and study filters to use the same structure in frontend and backend, simplifying the code here.  We then add new collision filters as specified in #874 , and update tests accordingly.

As part of this, the common date filter can now be _one-sided_: that is, you can specify "from this date" without specifying "to this date", and vice-versa.  This should make it easier to query for "everything since ...".

Some of these changes will require follow-up data pipeline work.

# Tests
Ran `npm run ci:jest-coverage`, which includes several new tests.  Tested quickly in frontend.


candu (commented):
LGTM

PR #878, opened 2021-04-27
Filter, search, sort tweaks in Track Requests

### LINKED ISSUE #871: Multi-word search terms not working in Track Requests
**What Happened**
From [user testing spreadsheet (internal only)](https://docs.google.com/spreadsheets/d/1ixtkOYs1RLO1DKO-7-_6MpnfNhEvSgRt70JUIstI6N0/edit#gid=1768524490):

> Only works when typing in the first letter of "Turning Movement Count" and "Multiple Locations", and the word "Ped" of "Ped Delay and Classification", and the word "Speed" of "Speed / Volume ATR". Does not work when full word is typed out, and if there are spaces between words.

**What Should Happen**
This is...a good question!  There are a few possible resolutions:

- decide that the current behaviour is OK, and leave it;
- remove some of the search columns from the list (i.e. Type, Assigned, Status), and expect users to use filters instead;
- fix these search columns (see Additional Notes below).

**To Reproduce**
Steps to reproduce the bug:
1. Log in as any user with `STUDY_REQUESTS` permissions;
2. Go to "Track Requests" using the left navbar;
3. Enter "Turning" in the search bar.

At this point, you'd expect Turning Movement Counts (TMCs) to come up in search results, but they don't.

**Additional Notes**
The underlying issue is that, in database, we use enum codes from `StudyType`, `StudyRequestAssignee`, and `StudyRequestStatus` - not the full text.  Moreover, these enum codes aren't kept in the database, but are instead listed in code (see `@/lib/Constants`):

```js
StudyType.init({
  // other keys...
  TMC: {
    label: 'Turning Movement Count',
    // other properties...
  },
});
```

Here the `label` is what shows up in the frontend, but it's stored in the database as `TMC`.

On top of this: to facilitate searching, we've introduced the table `study_request_items`, which combines non-bulk requests with aggregated values across bulk requests to form an index structure to support filter, search, and sort.  This is what allows us to treat bulk and non-bulk requests identically in the table.

This means that, while we *could* just store the `label` values in `study_request_items`...what if they change?  (For instance: we're already talking about re-labelling `ATR_VOLUME_BICYCLE`.)  That would mean rebuilding all of `study_request_items`, adding more maintenance behind the scenes (and more potential for surprises if that maintenance doesn't happen!)

Personally, I'd recommend we *don't* offer search on status, assignee, or type: these are much, much easier to handle as filters.  (That would leave ID, location, and requester as the only searchable columns.)  @shinechaudhuri @mkewins - thoughts?

### LINKED ISSUE #872: Sorting issues in Track Requests
**What Happened**
From our [QA testing spreadsheet (internal only)](https://docs.google.com/spreadsheets/d/1ixtkOYs1RLO1DKO-7-_6MpnfNhEvSgRt70JUIstI6N0/edit#gid=1768524490), some notes about sorting missing from the study type and status columns:

> can't sort by study type anymore (in dev / qa)? can currently sort by study type in prod

> removed from this version? at the same time, i think this is more useful for filter than sort

**What Should Happen**
This is...a good question!  As noted, status is a better fit for filtering than sorting; the same may be true of study type.  This leaves us with a couple of options:

- decide that we're not going to offer sorting on one (or both) of these columns;
- decide what sorting should mean on these columns, especially where bulk requests are concerned, and implement that.

**To Reproduce**
Steps to reproduce the bug:
1. Log in as any user with `STUDY_REQUESTS` permissions;
2. Click on "Track Requests" in the left navbar.

At this point, the table shows up, but you cannot sort on study type or status.

**Additional Notes**
Needs input from @shinechaudhuri @mkewins on the right path forward.

The main challenges here are maintaining `study_request_items` (see Additional Notes in #871 for a fuller explanation) and figuring out what to do with bulk requests.

IMHO we should just not support sorting on these columns - if the intent is to find a specific type of study, that's what filters are for!

candu:
# Issue Addressed
This PR closes #871 and closes #872 .

# Description
We remove full-text search support for Assignee, Status, and Type columns, and add filter support for Date Requested and Date Expected columns.

# Tests
Tested quickly in frontend.

### DIFF
--- a/lib/db/filters/StudyRequestFiltersSql.js
+++ b/lib/db/filters/StudyRequestFiltersSql.js
@@ -56,9 +56,6 @@ function getTextSearchFilter(column) {
  */
 function getStudyRequestFiltersSearch(column, query) {
   const filtersSearch = [];
-  if (column === null || column === 'ASSIGNED_TO') {
-    filtersSearch.push(getTextSearchFilter('"searchAssignedTo"'));
-  }
   if (column === null || column === 'ID') {
     // For ID searches, we expect an exact match.
     if (Number.isNaN(parseInt(query, 10))) {
@@ -72,12 +69,6 @@ function getStudyRequestFiltersSearch(column, query) {
   }
   if (column === null || column === 'REQUESTER') {
     filtersSearch.push(getTextSearchFilter('"searchRequester"'));
-  }
-  if (column === null || column === 'STATUS') {
-    filtersSearch.push(getTextSearchFilter('"searchStatus"'));
-  }
-  if (column === null || column === 'STUDY_TYPE') {
-    filtersSearch.push(getTextSearchFilter('"searchStudyType"'));
   }
   return filtersSearch;
 }
--- a/lib/db/filters/StudyRequestFiltersSql.js
+++ b/lib/db/filters/StudyRequestFiltersSql.js
@@ -87,6 +87,10 @@ function getStudyRequestFilters(studyRequestQuery, user) {
   const {
     assignees,
     column,
+    createdAtEnd,
+    createdAtStart,
+    dueDateEnd,
+    dueDateStart,
     query,
     statuses,
     studyTypes,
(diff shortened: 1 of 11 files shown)


lib/db/filters/StudyRequestFiltersSql.js
```diff
@@ -56,9 +56,6 @@ function getTextSearchFilter(column) {
  */
 function getStudyRequestFiltersSearch(column, query) {
   const filtersSearch = [];
-  if (column === null || column === 'ASSIGNED_TO') {
-    filtersSearch.push(getTextSearchFilter('"searchAssignedTo"'));
```
candu, line 60: this does technically mean we don't need those columns anymore, but we can clean them up later (e.g. if we need to squeeze a bit more performance out of Track Requests at some point).

candu (commented):
LGTM

PR #879, opened 2021-04-28
Download paginated requests: selected, all

### LINKED ISSUE #866: Selection / export for paginated study requests
**Description**
One remaining piece of study request pagination is the interaction between selection, bulk request updates, and export.  In this task, we decide on a way to handle that, then implement it.

**Acceptance Criteria**
- [ ] decide on a path forward here (possibly based on usability feedback);
- [ ] implement it (exact steps TBD).

**Additional Notes**
Adding `reporter` tag here, as one way to handle exporting would be to make this a new report type!

From Slack thread by @candu :

> realized I should mention one question that came out of the study request pagination: what do we do with selection / export?
> 
> there are two main types of bulk selection tasks here: status / assignment updates, and export.  one thought I had is that only being able to select within a page might actually be OK for status / assignment updates - that way, you're not surprised by updates to things you can't see in the table anymore.  OTOH, only being able to export a page at a time is likely going to be annoying - so a possible approach here is to decouple export from selection, and just change it to "export all matching filters" or something like that.
>
> this might be a useful question for usability testing?  mainly this is my own best guess on what makes sense here - it seems reasonable, but hasn't been checked with users in any way yet.

candu:
# Issue Addressed
This PR closes #866 .

# Description
We introduce two new report types: `TRACK_REQUESTS` (download all) and `TRACK_REQUESTS_SELECTED` (download selected).  We then implement those two report types as new report modules that subclass `ReportBaseTrackRequests`.  In the process, we move `RequestItemExport` logic out of the frontend and into the backend, reusing it to form the core of these reports.

# Tests
Tested quickly in frontend - this could likely use more unit testing, but it's also important to ensure we can get it through QA testing before deployment.

### DIFF
--- a/lib/io/storage/StoragePath.js
+++ b/lib/io/storage/StoragePath.js
@@ -300,7 +300,7 @@ class StoragePath {
     const partTimestamp = StoragePath.getTimestamp();
     const partOptionsHash = StoragePath.getOptionsHash(options);
     const { extension } = format;
-    const key = `${partReportType}_${partTimestamp}_${partOptionsHash}.${extension}`;
+    const key = `${partReportType}_${partTimestamp}${partOptionsHash}.${extension}`;
     return { namespace: StoragePath.NAMESPACE_REPORTS_STUDY_REQUEST, key };
   }
 
--- a/lib/io/storage/StoragePath.js
+++ b/lib/io/storage/StoragePath.js
@@ -311,7 +311,7 @@ class StoragePath {
    */
   static async forReport(report) {
     const { type } = report;
-    if (type === ReportType.TRACK_REQUESTS) {
+    if (type === ReportType.TRACK_REQUESTS || type === ReportType.TRACK_REQUESTS_SELECTED) {
       return StoragePath.forStudyRequestReport(report);
     }
     if (type.reportExportMode === ReportExportMode.COLLISIONS) {
--- a/lib/io/storage/StoragePath.js
+++ b/lib/io/storage/StoragePath.js
@@ -5,6 +5,7 @@ import {
   CentrelineType,
   LocationSelectionType,
   ReportExportMode,
+  ReportType,
 } from '@/lib/Constants';
 import { mapBy } from '@/lib/MapUtils';
 import ObjectUtils from '@/lib/ObjectUtils';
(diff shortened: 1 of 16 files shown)


lib/io/storage/StoragePath.js
```diff
@@ -282,13 +293,27 @@ class StoragePath {
...
   }
 
+  static async forStudyRequestReport(report) {
+    const { type, format, ...options } = report;
+
+    const partReportType = StoragePath.getReportType(type);
+    const partTimestamp = StoragePath.getTimestamp();
+    const partOptionsHash = StoragePath.getOptionsHash(options);
+    const { extension } = format;
+    const key = `${partReportType}_${partTimestamp}_${partOptionsHash}.${extension}`;
```
candu, line 303: `getOptionsHash` already prepends an underscore - we can drop the last underscore here.

lib/reports/ReportTrackRequests.js
```diff
@@ -0,0 +1,81 @@
+/* eslint-disable class-methods-use-this */
+import { ReportType } from '@/lib/Constants';
+import { mapBy } from '@/lib/MapUtils';
+import StudyRequestDAO from '@/lib/db/StudyRequestDAO';
+import StudyRequestBulkDAO from '@/lib/db/StudyRequestBulkDAO';
+import StudyRequestItemDAO from '@/lib/db/StudyRequestItemDAO';
+import CompositeId from '@/lib/io/CompositeId';
+import ReportBaseTrackRequests from '@/lib/reports/ReportBaseTrackRequests';
+
+const CHUNK_SIZE = Math.min(100, CompositeId.MAX_FEATURES);
```
candu, line 10: get rid of this and use the value from `ReportBaseTrackRequests`

lib/reports/ReportTrackRequests.js
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Subclass of {@link ReportBaseTrackRequests} for exporting study requests from Track Requests.
```
candu, line 13: update this comment to reflect purpose

lib/reports/ReportTrackRequestsSelected.js
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable class-methods-use-this */
+import { ReportType } from '@/lib/Constants';
+import StudyRequestDAO from '@/lib/db/StudyRequestDAO';
+import StudyRequestBulkDAO from '@/lib/db/StudyRequestBulkDAO';
+import { InvalidReportIdError } from '@/lib/error/MoveErrors';
+import ReportBaseTrackRequests from '@/lib/reports/ReportBaseTrackRequests';
+
+/**
+ * Subclass of {@link ReportBase} for exporting study requests from Track Requests.
```
candu, line 9: update this comment to reflect purpose

candu (commented):
mostly good, a few small nits to fix

candu (commented):
LGTM

PR #894, opened 2021-05-03
Collision filters UI: new filters, contextual help, filter chip grouping

### LINKED ISSUE #885: Address AND / OR confusion in Global Filters
**Description**
From usability testing on Global Filters, there's some confusion around whether filters are combined with AND vs. OR.  (The way it works functionally is reasonable, but unclear to users.)  In addition, there's some confusion around what different filters mean, and what the individual values apply to.

This task aims to address these points of confusion.

**Acceptance Criteria**
- [x] filters within a category are grouped into a single chip, similar to Track Requests;
- [x] emphasis areas / details have tooltips explaining how these are computed;
- [x] additional contextual help text is provided where needed.

**Additional Notes**
One concern here is keeping tooltip explanations of emphasis areas / details in sync with their actual backend implementation.  For now, it's probably enough to document somewhere (e.g. in the code itself) that both have to be updated together.

candu:
# Issue Addressed
This PR makes significant progress on #874 and closes #885 .

# Description
We add collision details, validated / not validated, MVCR available / missing, and vehicle type (`vehtype`) as new filters in `FcCollisionFilters`.  The `injury` and `vehtype` filter options are grouped to make these easier to use / understand, and we add contextual help at the field, group, and individual option levels for several collision filter categories.  Finally, we group filter chips for the same filter category into a single chip, similar to what was done for Track Requests.

# Tests
Tested quickly in frontend.


web/App.vue
```diff
@@ -148,6 +148,10 @@ export default {
   & .v-tooltip__content {
     background: rgba(33, 33, 33, 0.9);
     font-size: 1rem;
+
```
candu, line 151: we don't actually need this - it was added to support the link to Vision Zero materials, but the tooltip interaction is such that clicking on links in tooltips is hard anyways, so the link was ultimately removed from the tooltip copy.

candu (commented):
one small change requested, otherwise good

candu (commented):
LGTM

Issue #902, opened 2021-05-06
Definition of school children needs to be updated


candu:
**What Happened**
As per discussion with the City's Vision Zero team (thanks @mkewins for pushing that conversation forward!):

> Those were the criteria that we used as part of the original VZ Road Safety Plan back in 2016.  However, the definition for school children has since been updated and is now based only on the following:
> - InvolveType = Pedestrian, Cyclist, Cyclist Passenger, In-Line Skater, Wheelchair
> - Age = 4 to 19

As noted, we're using an outdated definition of "school children" to mark collisions that involve them.

**What Should Happen**
We should use the new definition noted above.  In particular, this definition differs from the one we're using as follows:

- it adds "In-Line Skater" and "Wheelchair" as active mobility options here;
- it removes time / date requirements that the collision occur on school days during school commute or lunch hours.

**To Reproduce**
Steps to reproduce the bug:
1. Click "Add" on the filter box;
2. Open the "Collisions" section;
3. Find "Vision Zero Emphasis Areas" and roll over the contextual help beside "School Children".

At this point, you'll see that we're using the older definition.

**Additional Notes**
The _real_ issue here is in our data pipelines - we need to update _that_ definition, and then we should update the contextual help to match.

candu:
Closed by two PRs (above).

PR #907, opened 2021-05-06
Update school children definition, use Vision Zero flags in popups

candu:
# Issue Addressed
This PR makes progress on #902 .

# Description
We update contextual help for Global Filters to match updates to the definition of school children (in the Vision Zero emphasis area sense) in our collision normalization data pipelines.

# Tests
Tested quickly in frontend.


candu (commented):
LGTM

PR #910, opened 2021-05-07
Update tests to account for post-conflation update dev dataset changes

### LINKED ISSUE #905: Previously loaded ATR not visible in MOVE
**What Happened**
From a user bug report:

> Some counts are not showing up in MOVE even though I loaded several days ago: ATR on blythewood between strathgowan cres and blythe hill road

Right now we run FLOW replication weekly, so this data should have appeared by now (the bug report is from Apr. 14, 2021) - but it hasn't.

**What Should Happen**
Data should be available in MOVE at most 1 week after being loaded in FLOW.

**To Reproduce**
Steps to reproduce the bug:
1. Go to [Blythwood Rd: near Blyth Hill Rd](https://move.intra.prod-toronto.ca/view/location/s1:A3J5SA/POINTS).

At this point, you can see that a Speed / Volume ATR has been requested, but no data is available.

**Screenshots**
Blythwood Rd: near Blyth Hill Rd - study requested, but no data yet

**Additional Notes**
We'll need to investigate the whole chain here, right back to the Oracle database that backs FLOW: loading, replication, or conflation could have failed.  If that's the case, we'll also potentially need better monitoring here.

To fully investigate, we may need to know the underlying arterycode.

candu:
# Issue Addressed
This PR closes #905 .

# Description
This follows up on a PR in `bdit_move_etl` to add a couple of double-link arterycode conflation cases - just need to quickly update a couple of tests to account for newly conflated counts.

# Tests
Ran `npm run ci:jest-coverage`.

### DIFF
--- a/tests/jest/api/StudyController.spec.js
+++ b/tests/jest/api/StudyController.spec.js
@@ -419,7 +419,7 @@ test('StudyController.getStudiesByCentrelineSummary', async () => {
   data = { s1 };
   response = await client.fetch('/studies/byCentreline/summary', { data });
   expect(response.statusCode).toBe(HttpStatus.OK.statusCode);
-  expectNumPerCategoryStudy(response.result, [[3633, 'RESCU'], [2, 'TMC']]);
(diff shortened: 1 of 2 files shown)


candu (commented):
LGTM

PR #923, opened 2021-05-12
Handle empty and stopwords-only queries without HTTP 500'ing

### LINKED ISSUE #531: Location query with only stopwords results in HTTP 500
**What Happened**
Searching for a location query that only contains stopwords (e.g. "the", "and", "and the", etc.) results in an HTTP 500 error.

**What Should Happen**
These queries should at least not crash, and would ideally work as normal fuzzy prefix queries.

**To Reproduce**
Steps to reproduce the bug:
1. Type "and" in the search bar
2. Open the browser's developer console

At this point, you should see an HTTP 500 error from `GET /locations/suggest`.

**Additional Notes**
The error in question is the result of `plain_totsquery()`'s stripping of stop words:

```sql
select plainto_tsquery('and');
NOTICE:  text-search query contains only stop words or doesn't contain lexemes, ignored
 plainto_tsquery 
-----------------
```

When we use this to build a prefix query for location search, it results in the invalid `tsquery` string `':*'`:

```sql
select (plainto_tsquery('and')::text || ':*')::tsquery;
NOTICE:  text-search query contains only stop words or doesn't contain lexemes, ignored
ERROR:  syntax error in tsquery: ":*"
```

Fully addressing this might involve [configuring a custom stopwords list](https://www.postgresql.org/docs/9.6/textsearch-dictionaries.html#TEXTSEARCH-STOPWORDS), or even parsing and normalizing the query in the MOVE web application server itself.

candu:
# Issue Addressed
This PR closes #531 .

# Description
Now runs a separate `plainto_tsquery()` SQL query before the exact / approximate matching SQL queries, to double-check whether this normalized version of the query is empty or not.

# Tests
Added test cases to `LocationSearchDAO.spec.js`.  Tested quickly in frontend as well.

### DIFF
--- a/lib/db/LocationSearchDAO.js
+++ b/lib/db/LocationSearchDAO.js
@@ -66,6 +66,12 @@ WHERE arterycode = $(arterycode)`;
    * most `limit` results
    */
   static async intersectionSuggestions(query, limit) {
+    const sqlQuery = 'SELECT plainto_tsquery($(query))::text AS "queryNormalized"';
+    const { queryNormalized } = await db.one(sqlQuery, { query });
+    if (queryNormalized === '') {
+      return [];
+    }
+
     /*
      * For exact matches, we use `tsvector` and `tsquery` full-text support together with
      * these features:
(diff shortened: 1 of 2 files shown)


candu (commented):
LGTM

Issue #884, opened 2021-04-29
Duplicate study request (both bulk and non-bulk)


candu:
**Description**
For before / after studies, it's useful to be able to duplicate previously-submitted requests.  This task aims to implement that.

**Acceptance Criteria**
- [ ] users with `STUDY_REQUESTS_ADMIN` scope:
  - [ ] can duplicate one or more selected study requests from Track Requests;
  - [ ] can duplicate a request (bulk or non-bulk) from its View Request page;
  - [ ] are prompted to create a new bulk request when duplicating a bulk request, or when duplicating multiple individual study requests.
- [ ] users without `STUDY_REQUESTS_ADMIN` scope cannot do this.

mkewins:
This feature request may be revisited in future, but is not a part of near-term priorities.

Issue #883, opened 2021-04-29
Additional study management controls


candu:
**Description**
Over the last while, we've received several feature requests around study management from @jwrcoleman and various users.  This task captures the lowest-hanging fruit among that feedback.

**Acceptance Criteria**
- [x] users with the given permissions:
  - [x] can change the requester of a study request... (`STUDY_REQUESTS_ADMIN`)
    - [x] the REST API permits this for non-bulk requests;
    - [x] the REST API permits this for bulk requests, in which case it changes the requester of all sub-requests as well;
    - [x] this is added as a test case in relevant test suites;
  - [x] can group several study requests into a single "project" (i.e. bulk request)... (`STUDY_REQUESTS`, `STUDY_REQUESTS_ADMIN`)
    - [x] a new REST API method for grouping requests into a bulk request is added (e.g. to `StudyRequestBulkController`);
    - [x] a new DAO method is added for this (e.g. in `StudyRequestBulkDAO`);
    - [x] this is added as a test case in relevant test suites;
    - [x] the frontend offers a way to do this (e.g. in Track Requests).
  - [x] can add one or more study requests to an existing "project" (i.e. bulk request)... (`STUDY_REQUESTS`, `STUDY_REQUESTS_ADMIN`)
    - [x] a new REST API method for adding requests to a bulk request is added;
    - [x] a new DAO method is added for this;
    - [x] this is added as a test case in relevant test suites;
    - [x] the frontend offers a way to do this (e.g. in Track Requests);
- [x] in all above cases, users without given permissions:
  - [x] cannot perform relevant REST API actions;
  - [x] cannot see relevant frontend UI in MOVE.

time permitting:
- [x] users with `STUDY_REQUESTS` scope:
  - [x] can change the location of a study request...
    - [x] the REST API permits this for requests;
    - [x] this is added as a test case in relevant test suites;
    - [x] the frontend offers a way to do this (e.g. in Edit Request).

**Additional Notes**
This might require a bit of design work from @shinechaudhuri - main concern here is interface clutter from adding these functions.  If we _do_ want to allow users to edit locations (unclear right now), we'll need design there as well.

This new functionality will likely mean no longer being able to show a "location selection" for bulk requests - we'll either have to get rid of that part of the bulk request, or change the map to just show a collection of points where study requests are taking place.

mkewins:
One other improvement data collection asked for:

> Can I edit an assigned request?

[...]

PR #930, opened 2021-05-18
Study request management: API changes

candu:
# Issue Addressed
This PR makes progress on #883 .

# Description
We make several changes to simplify the data model of `study_requests` and `study_requests_bulk`, and update `study_request_items` accordingly.  This leads to several downstream changes in DAOs, validation models, test data generators, etc.

We then introduce new REST API endpoints:

- `POST /requests/study/{id}/copy`: copies the given request;
- `POST /requests/study/bulk/{id}/copy`: copies the given project, including all requests within it;
- `POST /requests/study/bulk/fromRequests`: creates a new project from the given existing requests;
- `PUT /requests/study/bulk/{id}/requests`: adds the given requests into the given project;
- `DELETE /requests/study/bulk/requests`: removes the given requests from their projects.

These allow us to handle duplication functionality from #884 , as well as the added management operations from #883 .

We also update `StudyRequestPermissions` - it is now possible for users with `STUDY_REQUESTS_ADMIN` scope to change the requester of a request or project.

Finally, we add a bunch of testing around these new REST API endpoints and the database methods that back them, as well as the `StudyRequestItemDAO` entries created to index requests and projects.

# Tests
`npm run ci:jest-coverage`.  (Frontend testing coming in subsequent PRs; this is primarily a backend-focused PR.)


lib/Constants.js
```diff
@@ -985,7 +985,7 @@ StudyRequestStatus.init({
     color: 'statusAssigned',
     dataViewable: false,
     detailsIndex: 2,
-    editable: false,
+    editable: true,
```
candu, line 988: Add a comment to indicate that `editable` is now used to indicate whether a request is "closed".  (We used to have a separate `study_requests` column for this, but we don't need that anymore.)

scripts/db/schema-20.down.sql
```diff
@@ -0,0 +1,69 @@
+BEGIN;
+
+-- backward migration SQL goes here
```
candu, line 3: get rid of these comments

scripts/db/schema-20.up.sql
```diff
@@ -0,0 +1,54 @@
+BEGIN;
+
+-- forward migration SQL goes here
```
candu, line 3: same here

candu (commented):
couple of minor comment / docs nits, otherwise good

candu (commented):
LGTM

PR #931, opened 2021-05-20
Study management UI

candu:
# Issue Addressed
This PR makes further progress on #883 .

# Description
We get several pieces of the site working again after backend changes, and refactor `FcPaneMap` into a new `FcMap` component that doesn't use `vuex` state.  We then switch over several key parts of the site to use `FcMap`.  We also refactor our `vue-router` setup, moving View Data and New / Edit Request routes into separate layouts in preparation for further work.

# Tests
Tested quickly in frontend.


web/components/geo/map/FcMapPopupActionViewData.vue
```diff
@@ -0,0 +1,133 @@
+<template>
+  <FcButton
```
candu, line 2: bit of a problem here: if we hover over a school (or other non-selectable feature), the `<v-card-actions>` part of the popup shows up as an awkward thin grey bar at bottom of the popup.

maybe move `SELECTABLE_LAYERS` et al. into `FcMapPopup` and use it as part of the `v-if` condition on `<v-card-actions>`?

candu (commented):
mostly looks good, couple of small changes

web/components/geo/map/FcMapPopupActionViewData.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <FcButton
+    v-if="featureSelectable"
```
candu, line 3: oops, don't add this in

candu (commented):
one more small change

candu (commented):
LGTM

PR #939, opened 2021-05-28
Fix invage bug in ReportCollisionDirectory

### LINKED ISSUE #915: Age missing from the CSV download for Collision Directory Report
**What Happened**
For a multi-location selection, the age column is blank for all values in the Collision Directory Report CSV download.

**What Should Happen**
Age should be present.

**To Reproduce**
- v1.3.1 on dev
- Firefox browser
- Locations (1) Lawrence / Howden / Underwriters (2) Birchmount / Canadian; without selecting intersections / midblocks in between

Steps to reproduce the bug:
1. Create a multi-location selection with two or more locations.
2. Go to View Data
3. Click on "View Reports" under the Collisions section
4. Download a CSV of the Directory Report
5. See that the age column is blank

**Screenshots**
n/a

**Additional Notes**
Do NOT attach reports here.

candu:
# Issue Addressed
This PR closes #915 .

# Description
This was always showing up blank for CSV exports - turns out that we shouldn't use `this.getCollisionFactorCode` for `invage`, as it's not a coded field but rather a numeric one.

# Tests
Tested quickly in frontend, both before and after the fix (to repro, and then to verify that the fix had the intended effect.)

### DIFF
--- a/lib/reports/ReportCollisionDirectory.js
+++ b/lib/reports/ReportCollisionDirectory.js
@@ -83,7 +83,7 @@ class ReportCollisionDirectory extends ReportBaseCrash {
       drivact: this.getCollisionFactorCode('drivact', drivact),
       drivcond: this.getCollisionFactorCode('drivcond', drivcond),
       initdir: this.getCollisionFactorCode('initdir', initdir),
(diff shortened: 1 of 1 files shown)


candu (commented):
LGTM

Issue #1052, opened 2021-08-24
Infra: Relocate the MOVE Help Centre somewhere else


candu:
**Description**
We currently keep the MOVE Help Centre on our AWS ETL dev machine.  This task aims to move that to a separate hosting environment; the closer we can get AWS ETL to just being a simple Airflow machine, the easier it is to automate our Airflow setup.

**Acceptance Criteria**
- [x] find another place to host the MOVE Help Centre;
- [x] move it there;
- [x] redirect from old AWS ETL dev paths using HTTP 301;
- [ ] (optional) set up a one-click way of updating this.

**Additional Notes**
If we manage to get a server for this, it might also make sense to move other documentation there.  Ideally we'd be able to get a separate server for `gitlab-runner` operations as well...

candu:
Addressed this with a few PRs here and in `bdit_move_etl` - marking closed.

PR #1056, opened 2021-08-25
Change link to MOVE Help Centre...again

candu:
# Issue Addressed
This PR makes further progress on #1052.

# Description
It turns out that the previous link is blocked by the corporate proxy, so I'm changing it to something that should not be.

# Tests
Tested while on VPN in an incognito window, to ensure that this works as expected for users who are a) on the City network and b) not part of our wiki.


candu (commented):
LGTM

Issue #968, opened 2021-06-14
Upgrade Vue, Vuetify


candu:
**Description**
MOVE uses Vue v2, but Vue is now on v3.  Vuetify (which we use for Material Design components) has a [Q3 2021 release target](https://vuetifyjs.com/en/introduction/roadmap/) for v3, which will itself use Vue v3.

This task 

**Acceptance Criteria**
- [x] investigate feasibility of upgrading to Vue 3.x ahead of Vuetify 3.x release;
- [ ] expand set of tasks here based on investigation.

candu:
Vuetify 3.x is [currently in alpha](https://next.vuetifyjs.com/en/getting-started/installation/#how-to-report-an-issue-or-bug) - we should probably wait for something a _bit_ more stable before starting the Vuetify upgrade process.

leonL:
I'm tracking and scheduling dependency updates as part of the release process. Closing this ticket

PR #1058, opened 2021-08-30
Pin sass to 1.32.x to reduce DEPRECATION WARNING log spam

### LINKED ISSUE #991: Address large amount of DEPRECATION WARNING log spam in frontend builds
**Description**
Right now, each frontend build or `webpack-dev-server` startup is accompanied by a _lot_ of warning messages like the following:

```
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
```

This task aims to reduce this log spam.

**Acceptance Criteria**
- [x] find a way to reduce this log spam and implement it;
- [x] make sure frontend builds still work properly.

**Additional Notes**
This is all from [Vuetify](https://vuetifyjs.com/en/), which we use as a Material Design component library.  One possible approach to fixing this is mentioned in [this StackOverflow question](https://stackoverflow.com/questions/67631879/nuxtjs-vuetify-throwing-lots-of-using-for-division-is-deprecated-and-will-be), which also links to [this Vuetify issue](https://github.com/vuetifyjs/vuetify/issues/13694).

(TL;DR: this _might_ be solvable with a small change in `package.json`.)

candu:
# Issue Addressed
This PR closes #991 .

# Description
The longer-term fix is to upgrade Vuetify to a version that fixes this, but since we're waiting on the Vuetify 3 release to do that...this is an OK fix for now.  (See #968 for the Vuetify / Vue upgrade task.)

# Tests
Ran `ci:npm-outdated`, ensured that `sass` did not show up as an outdated package.  Ran frontend / backend, checked that site still loads in local dev, and that the frontend log output does not include loads of DEPRECATION WARNING messages.

### DIFF
--- a/.ncurc.json
+++ b/.ncurc.json
@@ -5,6 +5,7 @@
     "d3-axis",
     "d3-scale",
     "d3-selection",
+    "sass",
     "sass-loader"
   ]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -25946,9 +25946,9 @@
       "integrity": "sha512-YZo3K82SD7Riyi0E1EQPojLz7kpepnSQI9IyPbHHg1XXXevb5dJI7tpyN2ADxGcQbHG7vcyRHk0cbwqcQriUtg=="
     },
     "sass": {
-      "version": "1.38.1",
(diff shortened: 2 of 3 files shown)


candu (commented):
LGTM

PR #1059, opened 2021-08-30
Show percentiles, mean to one decimal place in Speed Percentile report

### LINKED ISSUE #974: Show one decimal place on summary stats for speed percentile reports
**What Happened**
An ATP user mentioned that the 46th percentile which is the median is the same value as the mean. He said he needs to know the mean, whether it's trending higher or lower and is expecting a decimal point. Are these rounded? Wanted to investigate this

candu:
# Issue Addressed
This PR closes #974 .

# Description
Adding a bit more precision here, as per user feedback.

# Tests
Updated `ReportSpeedPercentile.spec.js` and ran this.  Tested in local development using frontend, backend, and reporter.


candu (commented):
LGTM

PR #1066, opened 2021-09-28
1027 map hidden bug

### LINKED ISSUE #1027: View Reports: CSS bug prevents map from showing at top when drawer is open
**What Happened**
In View Reports, when the report drawer is open, the map is not visible at all at top, and the "Collapse Page" button shows up over top of the drawer.

**What Should Happen**
A bit of map (~60px) should be visible at top, and the "Collapse Page" button should be over top of the map at center.

**To Reproduce**
Steps to reproduce the bug:
1. Go to any location with collisions or studies
2. Click on "View Reports"

**Screenshots**
Here's View Reports as it currently is:

image

**Additional Notes**
You can see the intended behaviour in [(DC) Classification ATR (internal-only)](https://www.figma.com/file/iDmHumIVsgTK2wpi949gSt/(DC)-Classification-ATR?node-id=0%3A1) on Figma.

leonL:
# Issue Addressed
This PR closes #1027 

# Description
Fixed the bug by changing the flex-basis rule on the map and drawer divs from _0_ to _auto_ so that explicit height values are respected when the divs are rendered

# Tests
Open the data drawer in a vertical and expanded state to verify that the underlying map is still visible at the top of the page

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5682,32 +5682,6 @@
           "integrity": "sha512-dOy+3AuW3a2wNbZHIuMZpTcgjGuLU/uBL/ubcZF9OXbDo8ff4O8yVp5Bf0efS8uEoYo5q4Fx7dY9OgQGXgAsQA==",
           "dev": true
         },
-        "emojis-list": {
-          "version": "3.0.0",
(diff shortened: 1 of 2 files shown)


candu (commented):
LGTM

PR #1068, opened 2021-10-04
skip tests failing due to variance in sample data

leonL:
# Description
Temporarily skip running automated tests that are failing in local dev due to variance in the sample data used on the test DB


candu (approved):
LGTM - we can also go over the rest of the tests that depend on live data in today's pairing.

PR #1123, opened 2024-03-11
Move 254: Sorting collision report data by date

lowsound42:
# Issue Addressed
This PR closes [MOVE-254](https://move-toronto.atlassian.net/browse/MOVE-254)

image

# Description
Collision Directory Reports were being populated in ascending order by date, while users are most likely looking to first see more recent data. This PR adds a simple sort to the CollisionDAO that ensures that data is first sorted by date (in descending order) before being sent to the report creator.

An additional issue was that some dates seemed out of order (i.e. not exactly in ascending order). This issue is also likely addressed by the same fix. Previously data was sorted by collision_id only.  collision_id is a string and will not ensure proper sorting. As of this change, the final sort is done on the date field.

# Tests
I haven't stepped into the project tests just yet. Everything was manually verified. 

- Navigate to the MOVE app
- Select an intersection
- View reports of collisions
- Verify that the collision report is displayed in descending order (by date)


spencermountain (approved):
good eye!

ryanpdaley (approved):
Looks good to me!

PR #1139, opened 2024-03-18
MOVE-488 - [Emails] Requestors should be notified when request needs clarification

ryanpdaley:
# Issue Addressed
This PR closes #[MOVE-488](https://move-toronto.atlassian.net/browse/MOVE-488)
# Description
Currently the requestor of a study does not receive an email when the request status is changed to `Needs Clarification`. This ticket changes this by sending an email to the requestor for both bulk study requests, and single requests on status change to `Needs Clarification`

# Tests
I've updated and passed all tests. I think we'll need to verify the functionality in dev - I can see the frontend changes, but I can't validate the emails being sent / received from local.


spencermountain (approved):
nice one Ryan!

lowsound42 (approved):
🤘

PR #1140, opened 2024-03-19
Move 1105: Speed percentile report headers not matching (csv vs pdf)

lowsound42:
# Issue Addressed
This PR addresses [MOVE-1105](https://move-toronto.atlassian.net/browse/MOVE-1105)

# Description
Speed percentile reports were generating different headers for the csv and pdf downloads. The issue was that the csv generator method was adding 1 to the low end of the range for each percentile (due to a change in the constants file not being propagated down to this method).

image

This has been changed and a corresponding test to see if the generated columns correspond to the expected values has been added as well.

# Tests
Tested locally on multiple speed percentile reports to see if csv and pdf download headers match, no changes to other parts of the files.
Test added to the test suite.


ryanpdaley (approved):
Nice - what an odd issue. haha

PR #1149, opened 2024-03-28
MOVE-754: Add pagination for admin console

imadsyed333:
# Issue Addressed
This PR closes MOVE-754

# Description
Added pagination to the User Permissions tab of the Admin Console, with a limit of 50 users per page, and stylized to resemble the Track Requests page.

Before:
image

After:
image

# Tests
Tested in MOVE local using Google Chrome and Microsoft Edge, by switching between pages with various page limits, and tested on various window sizes.

### DIFF
--- a/lib/controller/UserController.js
+++ b/lib/controller/UserController.js
@@ -21,18 +21,18 @@ const UserController = [];
  * @name getUsers
  * @type {Hapi.ServerRoute}
  */
-UserController.push({
-  method: 'GET',
-  path: '/users',
-  options: {
-    description: 'Get all users',
-    response: {
-      schema: Joi.array().items(User.read),
-    },
-    tags: ['api'],
-  },
-  handler: async () => UserDAO.all(),
-});
+// UserController.push({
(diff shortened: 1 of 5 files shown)


spencermountain (approved):
this is so good! nice work Imad

lib/controller/UserController.js
```diff
@@ -21,17 +21,58 @@ const UserController = [];
  * @name getUsers
  * @type {Hapi.ServerRoute}
  */
+// UserController.push({
```
ryanpdaley, line 24: @imadsyed333 - can you delete this if we don't need it anymore

ryanpdaley (commented):
Looks good to me! 
Two minor things:
- See comment about deleting the commented out block
- When you are going to merge, can you 'Squash and Merge' - I kept forgetting to do this, but it make this one commit to master rather than 13

lib/controller/UserController.js
imadsyed333: Sure!

PR #1150, opened 2024-04-02
Move-1113

lowsound42:
# Issue Addressed
This PR closes [MOVE-1113](https://move-toronto.atlassian.net/browse/MOVE-1113)

# Description
This PR addresses two issues.

1. When a collision includes invalid data (for example, negative actual_speed on one or more involved parties), the app did not handle the error on the server side at all. A try/catch block was added in the collision controller (for both, the collision factor and collision endpoints) and a Boom error is generated if the retrieval fails. Similarly, the collision directory report depended on the successful retrieval of a collision, so the report controller also now has a try catch block added to the report endpoint to catch this and generate a Boom error.

2. Boom errors are now handled globally via the MoveServer. Using the 'onPreResponse' lifecycle hook, we can intercept a boom response, log the data to the new app_logs table, and then return the response to the user.

A new table called app_logs has been added to the DB and the migration scripts are included, devs will need to run the update-db script to update to version 25 of the database.

# Tests
The controller for collision controllers has been tested to ensure that a failed retrieval throws t he correct Boom error. The report controller is not so easy to test because every report request is different and so mocking the tests is not easy. I notice that no existing tests have been written for the report controller and this probably something we need to look into.

Next steps -> other controllers will need to be examined for appropriate error handling and if required, try/catches need to be added where a DAO is accessed and an appropriate Boom error fired in case of failure.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -204,6 +204,9 @@ HttpStatus.init({
   NOT_FOUND: {
     statusCode: 404,
   },
+  INTERNAL_SERVER_ERROR: {
+    statusCode: 500,
+  },
 });
 
 /**
--- a/lib/controller/CollisionController.js
+++ b/lib/controller/CollisionController.js
@@ -1,5 +1,7 @@
 import Boom from '@hapi/boom';
-
+import {
+  HttpStatus,
+} from '@/lib/Constants';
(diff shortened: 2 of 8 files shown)


spencermountain (approved):
🔥

PR #1151, opened 2024-04-02
MOVE-1000 - Peak Hour calculation not updated in the Peak Hour Factor Report

ryanpdaley:
# Issue Addressed
This PR closes [MOVE-1000](https://move-toronto.atlassian.net/browse/MOVE-1000)

**Before:**
before

**After:**
after

# Description
Previously we were only looking at the hour value for finding and reporting on the peak hours. This updates the methods to use the minutes also since the AM and PM peak ranges include quarter-hour values (15min, 30min, 45min).

Also unified all of these ranges in the Constants, and updated the reports that use these values.

# Tests
Updated and passed all tests.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -1249,6 +1249,27 @@ const StudyRequestChanges = {
   },
 };
 
+const REPORT_CONSTANTS = {
+  COUNT_INTERVAL_DURATION: { minutes: 15 },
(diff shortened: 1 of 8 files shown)


lowsound42 (approved):
Time is an illusion. Lunchtime doubly so.

spencermountain (approved):

PR #1152, opened 2024-04-03
Move 515

spencermountain:
# Issue Addressed
This PR closes [MOVE-515](https://move-toronto.atlassian.net/browse/MOVE-515) (new callout in reporter html) and [MOVE-1124](https://move-toronto.atlassian.net/browse/MOVE-1124) (show one tooltip at a time). It also makes some minor changes to the legend, from [MOVE-1109](https://move-toronto.atlassian.net/browse/MOVE-1109).

# Description
* hides legend when left sidebar opens
* hides legend on minimum height, not just min width
* puts map tooltip above legend (via  `z-index` )
* adds `user-select: none` to missing menu and button elements

here is the new callout button:
image
it will only show when there is the width for it

Accidentally added Omar's clustering commit from https://github.com/CityofToronto/bdit_flashcrow/compare/master...MOVE-479
I reverted it, so the PR is clean, and I believe it will prevent any problems when that branch merges

### DIFF
--- a/lib/geo/GeoStyle.js
+++ b/lib/geo/GeoStyle.js
@@ -226,6 +226,9 @@ function addGeoJsonSource(style, id) {
   style.sources[id] = {
     type: 'geojson',
     data: CACHE_GEOJSON[id],
+    cluster: true,
+    clusterMaxZoom: 15, // Max zoom to cluster points on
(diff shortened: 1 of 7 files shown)


lowsound42 (approved):
💯

PR #1164, opened 2024-04-10
(Update) MOVE-1100 - Export comments with request queue CSV export

ryanpdaley:
# Issue Addressed
This PR closes [MOVE-1100](https://move-toronto.atlassian.net/browse/MOVE-1100)

# Description
Previous PR had a race collision that I got stuck trying to work around... so I rewrote this function, and mirrored it on the existing get functions... and it works!

# Tests
I tested in local with reports that had 1 comment, no comments, huge comments, deleted comments and verified that the project requests view is still functional.


lowsound42 (approved):
"I'd have done it too if it wasn't for that pesky city developer"

- the race condition probably

PR #1166, opened 2024-04-11
April feedback tweaks

spencermountain:
a variety of jira feedback tweaks, after maddy's vacation. 

# Description
* undo legend hide with sidebar
* alignment/padding tweaks to sidebar changes
* Hide top map element on Report view
* show pagination buttons in report-type tab elements
* move new callout button to it's own row
* only show Day-Picker when there is > 1 day to show
* disable map popups above zoom 12

may conflict some of Omar's upcoming work disabling report buttons. 😟


spencermountain:
image

web/components/reports/FcReportMetadata.vue
```diff
@@ -12,6 +13,22 @@
       </dd>
     </v-col>
   </v-row>
+
```
lowsound42, line 16: Just a heads up, I've created a new component called FcCallout that I basically copied from this callout you created @spencermountain. I needed to use it in two places so it made sense. You should be able to use it here as well.

lowsound42 (approved):

PR #1172, opened 2024-04-15
MOVE-1148: Add informative tooltips to location markers in project view

imadsyed333:
# Issue Addressed
This PR closes MOVE-1148

# Description
Added a tooltip to location markers in the map of MOVE Project View. Tooltip displays study request ID, type, location, and duration. If there are multiple study requests at a single location, the tooltip will display them in a list. Each request also has a button to take the user to the Request page of that request.

image

# Tests
Tested in MOVE local v1.11.1 in Chrome and Firefox.


web/components/geo/map/FcMapPopup.vue
```diff
@@ -81,6 +90,9 @@ export default {
     featureSelectable() {
       return SELECTABLE_LAYERS.includes(this.feature.layer.id);
     },
+    isStudyRequest() {
+      return !(this.feature.properties.studyRequests === undefined);
```
spencermountain, line 94: no biggie, i'd use `!==`

spencermountain (approved):
nice work!!

web/components/geo/map/FcMapPopup.vue
imadsyed333: I'm now left baffled as to why I didn't do that 😅

ryanpdaley (approved):
I think this looks okay! There is something weird going on with the package-lock.json needing to be committed... but I am assuming this had to do with a fix that you tried for the lining issue... we'll see 👀

imadsyed333:
> I think this looks okay! There is something weird going on with the package-lock.json needing to be committed... but I am assuming this had to do with a fix that you tried for the lining issue... we'll see 👀

Oh I think this might have been because I ran npm install to see if it would fix the linting issue (it didn't 😞 ).

PR #1176, opened 2024-04-17
Move-1136: Ripping out old MVCR stuff and replacing with a better (??) way

lowsound42:
# Issue Addressed
This PR closes [MOVE-1136](https://move-toronto.atlassian.net/browse/MOVE-1136)

# Description
The collision directory report view includes links to MVCR documents on rows (where available). The availability of these files and the data required to make the api call to retrieve them was being taken from data that was in the table. i.e. table columns were being used by the UI to construct strings that were then used to make api calls. Works okay but breaks if a new column is added or a column header is changed.

I added a property to the mvcr column (in addition to value and style) called mvcrDetails that is populated with the mvcr information on the server before being sent to the client. This way the cell has all the information it needs already embedded inside itself.

Additionally, were were doing some gymnastics to create a list of all mvcrs in a report to begin the bulk mvcr download process. This is still not a super elegant process but its a bit cleaner since we use the mvcrDetails objects from each row to create the list more efficiently.

# Tests

- Added data in the test cases to make sure that the new mvcrDetails data was being checked and tested for.
- Tested multiple reports across the map
- Tested tabulation report to make sure it was not breaking due to these changes.


spencermountain (approved):
i mvCr what you did there

ryanpdaley (approved):
Whoo.

PR #1181, opened 2024-04-22
MOVE-1146: Traffic Control Signal number is "None" in report headers

ryanpdaley:
# Issue Addressed
"This PR closes [MOVE-1146](https://move-toronto.atlassian.net/browse/MOVE-1146)

# Description
It seems from looking at the test cases for reports, that we used to get PX as part of the location description (ex.: _OVERLEA BLVD AT THORNCLIFFE PARK DR & E TCS (PX 679)_). This doesn't appear to be the case anymore, so we need to get it from the PoiDao.

image

# Tests
Updated the tests to mock the PoiDAO, all existing tests pass, and the PX value is appearing in the tests where 'None' was previously. Also, I left the old logic in as a fallback incase the PoiDAO doesn't return anything.


lowsound42 (approved):
makes sense 👍🏽

spencermountain (approved):

PR #1182, opened 2024-04-23
Move 1142: Awaiting progressCurrent in MVCR bulk download

lowsound42:
# Issue Addressed
This PR closes MOVE-1142

# Description
In the local environment, some bulk MVCR downloads were failing and left in an orphaned state.
This seems to be because the progressCurrent property was being updated in parallel with the status of the job, however it was not being awaited. As such, a race condition existed where it could be updated AFTER the job was completed leading to a broken but completed job.

# Tests
Tested locally on multiple MVCR downloads.


lib/jobs/JobRunnerCompressMvcrs.js
```diff
@@ -37,7 +37,7 @@ class JobRunnerCompressMvcrs extends JobRunnerBase {
           fileCompressed = true;
         }
 
-        this.incrProgressCurrent();
+        await this.incrProgressCurrent();
```
ryanpdaley, line 40: lol

ryanpdaley (approved):

spencermountain (approved):

PR #1199, opened 2024-04-29
change property from verified to validated

spencermountain:
# Issue Addressed
fixes validated/verified property in mulit-location sidebar

image


ryanpdaley:
👍

PR #1202, opened 2024-04-29
Move 1097: Persisting filters on refresh/login

lowsound42:
# Issue Addressed
This PR closes [MOVE-1097](https://move-toronto.atlassian.net/browse/MOVE-1097)

# Description
Filters applied by users did not generally persist, with the notable exception of the MVCR download flow. This was because the app was using the router to send the filters to the next screen after login (using that mvcr login button). This could not be used for logins via the actual login button since it lives outside of the router-view.

The new flow does away with the above. Now we set the filter state in session storage. On refresh, this is used to populate filters (whether logged in or not). Filters can be removed as normal.

# Tests
Added a test for new time formatter function that converts formatted time strings to luxon time

### DIFF
--- a/lib/time/TimeFormatters.js
+++ b/lib/time/TimeFormatters.js
@@ -1,4 +1,4 @@
-import { Info } from 'luxon';
+import { Info, DateTime } from 'luxon';
 
 import ArrayUtils from '@/lib/ArrayUtils';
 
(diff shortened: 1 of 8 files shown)


ryanpdaley (approved):
👍

spencermountain (approved):
👏

PR #1203, opened 2024-04-29
MOVE-487: Requestors need a better way to interact between projects and their constituent study requests

imadsyed333:
# Issue Addressed
This PR addresses MOVE-487

# Description
Improved the UI of the MOVE Project / Bulk Request View, added tooltips to map pins, and interactions between request table and map.
image

# Tests
Tested in MOVE v1.11.1 in Firefox

### DIFF
--- a/lib/geo/CentrelineUtils.js
+++ b/lib/geo/CentrelineUtils.js
@@ -222,19 +222,19 @@ function groupRequestsByLocation(locationsState) {
   const locationGroups = {};
   for (let i = 0; i < locationsState.length; i++) {
     const studyRequest = locationsState[i];
-    const { description } = studyRequest.location;
-    if (description in locationGroups) {
+    const { centrelineId } = studyRequest.location;
+    if (centrelineId in locationGroups) {
       const {
         requestId, requestType, numDays, requestHours,
       } = studyRequest.location;
-      locationGroups[description].location.studyRequests.push({
+      locationGroups[centrelineId].location.studyRequests.push({
         requestId, requestType, numDays, requestHours,
       });
     } else {
       const {
         requestId, requestType, numDays, requestHours, ...rest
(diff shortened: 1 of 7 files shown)


web/views/FcRequestStudyBulkView.vue
```diff
@@ -113,9 +129,8 @@ import { mapActions, mapMutations } from 'vuex';
 
 import { centrelineKey, ProjectMode, StudyRequestStatus } from '@/lib/Constants';
 import { getStudyRequestBulk } from '@/lib/api/WebApi';
-import { getStudyRequestLocation } from '@/lib/geo/CentrelineUtils';
+import { getStudyRequestInfo, groupRequestsByLocation } from '@/lib/geo/CentrelineUtils';
```
lowsound42, line 132: groupRequestsByLocation isn't being used anywhere in this file. Do we still need to import it?

lib/geo/CentrelineUtils.js
```diff
@@ -191,6 +193,64 @@ function getStudyRequestLocation(studyRequest, location) {
...
+  const { label } = studyRequest.studyType;
+  const { id, hours, daysOfWeek } = studyRequest;
+  if (location !== null) {
+    return {
+      ...location,
+      requestType: label,
+      requestId: id,
+      requestHours: (hours === null ? null : hours.description),
+      numDays: daysOfWeek.length,
+    }
```
lowsound42, line 209: Missing semicolon here is going to throw a linting error.

ryanpdaley (changes requested):
We'll need to get your linter working. I think the things that @lowsound42 has pointed out may block the project from building when it gets deployed... (and it means that nobody else will be able to commit changes without fixing these warnings/errors in their branches)

PR #1212, opened 2024-05-03
Float sidebar

spencermountain:
# Issue Addressed
This PR closes **MOVE-1173**  *Reports page: bring back map peek ability* - and addresses **MOVE-1176** *Webapp should be more responsive*

# Description
Absolutely positions `fc-drawer` component, so that both the sidebar and report-view float above the map.
This allows users to peek back at the map, when viewing a report. It also allows us to control the width of the sidebar, instead of fixing it at 50%. It also allows us to only use the sidebar height that we need.

* changes are duplicated in Study and Collision report pages
* adds some min-sizing to reduce jumping in the css transition
* adds some sidebar styling so it will behave in floating-mode
* adds tooltips for the new report buttons
* removes unused **vertical-mode** feature code
* removes old sidebar toggle button, which was prev absolutely positioned
* did not change map zoom (easetolocation), but may need to be nudged a wee bit

image

image

image

### DIFF
--- a/web/components/FcDrawerViewCollisionReports.vue
+++ b/web/components/FcDrawerViewCollisionReports.vue
@@ -16,7 +16,7 @@
       aria-label="Loading collision reports viewer" />
     <template v-else>
       <div>
-        <div class="align-center d-flex flex-grow-0 flex-shrink-0 px-3 pt-2">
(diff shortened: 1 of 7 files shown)


ryanpdaley (approved):
🎉 looks good!

lowsound42 (approved):
🆒

PR #1214, opened 2024-05-06
MOVE-1180: Collision Directory Report should not overflow page width

ryanpdaley:
# Issue Addressed
This PR closes [MOVE-1180](https://move-toronto.atlassian.net/browse/MOVE-1180)

# Description
Essentially this ticket is just trying to make the Collision Directory Report not go off the pdf page.

The things that I have changed:

- I versioned up pdfmake (the library we use for generating pdfs)
- I adjusted the margins a bit (went from l (large) to m (medium) and that seems to give us a bit more room. 
- I modified the column headers of the involved parties count
it was:

```
|   #    |  #  |    #    |
| Driver | Ped | Cyclist |
```
and I changed it to:

```
|  #   |  #  |  #  |
| Driv | Ped | Cyc |
```
The values within these columns are the counts for the involved parties, so I am assuming they are almost always <=9, but the headers are holding enough space for the columns to be a 6 (Driver) or 7 (Cyclist) digit number. 

# Tests
All test are passing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "module-alias": "^2.2.3",
     "mustache": "^4.2.0",
     "openid-client": "^4.9.1",
-    "pdfmake": "^0.2.7",
+    "pdfmake": "^0.2.10",
     "pg-boss": "^6.2.2",
     "pg-promise": "^10.15.4",
     "underscore": "^1.13.6",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
     "module-alias": "^2.2.3",
     "mustache": "^4.2.0",
     "openid-client": "^4.9.1",
-    "pdfmake": "^0.2.7",
+    "pdfmake": "^0.2.10",
```
lowsound42, line 64: 👀

lowsound42 (approved):

spencermountain (approved):

PR #1206, opened 2024-04-30
MOVE-487: Requestors need a better way to interact between projects and their constituent study requests

imadsyed333:
# Issue Addressed
This PR closes MOVE-487

# Description
UI overhaul of the MOVE Project / Bulk Request Page, tooltips for map pins, interactions between request table and map.

image

# Tests
Tested in MOVE v1.12.0 using Firefox

PR #1216, opened 2024-05-07
fix issue with multiple api calls on hover

lowsound42:
# Issue Addressed
This is an amendment to [MOVE-487](https://github.com/CityofToronto/bdit_flashcrow/pull/1206)

# Description
loadAsyncForFeatures being a computed property in the FcMapPopup component was resulting in multiple API calls going out for a single hover on/hover off event. This is likely not the way we want to be handling this. @imadsyed333, just a heads up that this will affect the project page popups.

### DIFF
--- a/web/components/geo/map/FcMapPopup.vue
+++ b/web/components/geo/map/FcMapPopup.vue
@@ -84,7 +84,6 @@ export default {
   },
   computed: {
     coordinates() {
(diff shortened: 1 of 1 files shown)


imadsyed333 (approved):
LGTM!

spencermountain (approved):

ryanpdaley (approved):

PR #1220, opened 2024-05-08
MOVE-1193: Display filters on collision directory report web page

imadsyed333:
# Issue Addressed
This PR closes MOVE-1193 and MOVE-1194

# Description
Added date range to the top right hand corner of the report. Also added a dropdown menu displaying collision filters and the number of them.

image

The same layout of collision and date filters is also present on the PDF export of the collision directory report.

image

# Tests
Tested in MOVE local v1.12.0 in Firefox and on various screen sizes.


web/components/reports/FcReportMetadata.vue
```diff
@@ -32,20 +49,86 @@
...
+    },
+  },
+  methods: {
+    readableFilters(filters) {
+      const readableFilters = [];
+      for (let i = 0; i < filters.length; i++) {
+        const { filter, label } = filters[i];
+        let newFilter = '';
+
+        if (filter === 'validated') {
```
spencermountain, line 95: i'd do something like
```js
let mapping = {
  daysOfWeek : 'Days of the week', 
  ....
}
newFilter = mapping[filter] || ''
```

spencermountain (approved):

web/components/reports/FcReportMetadata.vue
ryanpdaley: Are these available to import from lib/Constants.js? 
... if not, maybe a switch statement would organize it a bit more.

package.json
```diff
@@ -44,6 +44,7 @@
     "archiver": "^5.3.1",
     "aws-sdk": "^2.1393.0",
     "axios": "^0.22.0",
+    "caniuse-lite": "^1.0.30001616",
```
spencermountain, line 47: is this intentional? I welcome it if it can make the builds faster!

spencermountain (approved):
nice work!

package.json
imadsyed333: I don't know if this made the builds faster, but eslint wanted it, and I think it's made linting more responsive for me.

ryanpdaley (approved):

lowsound42 (approved):

PR #1221, opened 2024-05-08
MOVE-59: changes to multi location collision directory report

lowsound42:
# Issue Addressed
This PR addresses [MOVE-59](https://move-toronto.atlassian.net/browse/MOVE-59)

# Description
When a user selects multiple locations and then views the collision directory report, the current behaviour dumps all the locations in one report with no distinction between them. For example, this is the report for 3 separate locations:

image

The new behaviour adds a drop down at the top of the report that allows the user to pick which location they'd like to view a report for.
image
image
image

This allows for users to differentiate between locations and make more sense of the collision reports.

This does not change how pdf and csv exports are handled, they still dump all three locations into one report. That is a feature that needs to be worked on after this.

# Tests
Tested with multiple locations, locations with no collisions, mvcrs downloaded successfully as well.

### DIFF
--- a/lib/controller/JobController.js
+++ b/lib/controller/JobController.js
@@ -82,6 +82,7 @@ JobController.push({
         collisionQuery,
         reportFormat,
       );
+
       if (reports.length === 0) {
         return Boom.notFound('no reports for given filters');
       }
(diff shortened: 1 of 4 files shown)


spencermountain (approved):
this is cool!

ryanpdaley (approved):
looks neat!

PR #1224, opened 2024-05-13
Move 59: Multi location report collision (fix for deep link error)

lowsound42:
# Issue Addressed
This is a fix for an issue introduced by the last PR for MOVE-59. In that case, deeplinking to a multi location url would break the report. This was because of code added to the beforemount hook in the component that needed to be conditional based on whether there were preselected locations.

This is now fixed.

### DIFF
--- a/lib/Constants.js
+++ b/lib/Constants.js
@@ -439,7 +439,8 @@ class ReportDataType extends Enum {
   }
 }
 ReportDataType.init([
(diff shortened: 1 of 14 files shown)


spencermountain (approved):
i also changed this file very slightly in #1225 but it should auto-merge. LGTM!

ryanpdaley (approved):

PR #1230, opened 2024-05-14
Move 1149v3

imadsyed333:
# Issue Addressed
This PR closes MOVE-1149

# Description
Made map bigger, while also retaining important study request info.

image

Also swapped the placement of bulk request details.

image

# Tests
Tested in MOVE local v1.12.0 using Firefox on various screen sizes.


ryanpdaley (approved):
I'm going to assume this will work - it doesn't look like it will break anything other than maybe a layout / view 🤪

PR #1231, opened 2024-05-14
Move 1209: temp fix for speed percentile studies

lowsound42:
# Issue Addressed
This PR closes [MOVE-1209](https://move-toronto.atlassian.net/jira/software/projects/MOVE/boards/1?assignee=712020%3A76a74782-6a99-4d85-bbb2-b0e95058d6b0&selectedIssue=MOVE-1209)

# Description
After a recent change, speed percentile reports seemed to be broken and displaying only 0's and nulls.
This happened because we are in the process of bringing in a new study type (speed/volume/classification) and the conditional logic to render the report data was preemptively changed. 

For now, an OR operator is being used to check whether the study type is ATR_SVC or ATR_SPEED_VOLUME and rendering the data for either (which at the moment really is only ATR_SPEED_VOLUME).

# Tests
Reports rendering as expected.

### DIFF
--- a/lib/api/WebApi.js
+++ b/lib/api/WebApi.js
@@ -27,7 +27,6 @@ import {
   normalizeJobMetadatas,
 } from '@/lib/model/helpers/NormalizeUtils';
 import DateTime from '@/lib/time/DateTime';
-import config from '@/lib/config/MoveConfig';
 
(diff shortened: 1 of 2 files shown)


ryanpdaley (approved):
👍

PR #1234, opened 2024-05-14
Move 1149v4

imadsyed333:
# Issue Addressed
Part of MOVE-1194

# Description
You can now hover over any part of the row, and the corresponding map pin will be highlighted.

image

# Tests
Tested in MOVE local v1.12.0 using Firefox

### DIFF
--- a/web/components/FcDataTable.vue
+++ b/web/components/FcDataTable.vue
@@ -40,6 +40,7 @@
 </template>
 
 <script>
+import { mapMutations } from 'vuex';
 import { Ripple } from 'vuetify/lib/directives';
 
 import { KeyCode } from '@/lib/Constants';
@@ -182,6 +183,18 @@ export default {
         }
       });
     });
+
+    // Adds hover event listeners to table rows
+    const $tableRows = this.$el.querySelectorAll('tbody tr');
+    $tableRows.forEach(($tr) => {
+      $tr.addEventListener('mouseover', () => {
+        const requestId = parseInt($tr.children[1].innerText.trim(), 10);
+        this.setHoveredStudyRequest(requestId);
+      });
+      $tr.addEventListener('mouseleave', () => {
+        this.setHoveredStudyRequest(null);
+      });
+    });
   },
   methods: {
     customSort(items, sortBy, sortDesc) {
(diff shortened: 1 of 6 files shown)


lowsound42:
image

I think we have the same problem here with multiple calls being made @imadsyed333, presumably this is still coming from the watcher.

lowsound42 (changes requested):
Still seeing the same issues with repeated calls going to the API on hover requests.

lowsound42:
sorry to be nitpicky about this again 😬 , but it seems to be the watcher in the FcMapPopup again 
image

If you take the deep check out of there, the repeated calls stop. I actually think this was in the last PR not specifically this one, but it is dealing with the same issue.

imadsyed333:
> sorry to be nitpicky about this again 😬 , but it seems to be the watcher in the FcMapPopup again image
> 
> If you take the deep check out of there, the repeated calls stop. I actually think this was in the last PR not specifically this one, but it is dealing with the same issue.

@lowsound42 I think my latest change solves it (it has on my end). I noticed that featureKey was computed based off the layer name and the id of the selected / hovered feature. However, the map pins all had undefined id, so featureKey never changed whenever you clicked or hovered between map pins, so the popup content was not being rendered. I added some logic to provide the featureKey with the centrelineId instead, so featureKey now changes when moving between map pins, and the watcher will get activated appropriately.

lowsound42:
> > sorry to be nitpicky about this again 😬 , but it seems to be the watcher in the FcMapPopup again image
> > If you take the deep check out of there, the repeated calls stop. I actually think this was in the last PR not specifically this one, but it is dealing with the same issue.
> 
> @lowsound42 I think my latest change solves it (it has on my end). I noticed that featureKey was computed based off the layer name and the id of the selected / hovered feature. However, the map pins all had undefined id, so featureKey never changed whenever you clicked or hovered between map pins, so the popup content was not being rendered. I added some logic to provide the featureKey with the centrelineId instead, so featureKey now changes when moving between map pins, and the watcher will get activated appropriately.

amazing, yeah, this looks like it's working 👍🏽

lowsound42 (approved):

PR #1236, opened 2024-05-15
remove strict browserlist requirement

spencermountain:
# Issue Addressed
This PR closes MOVE-1200 - speed-up development script

Tried a bunch of different webpack updates and configs, and nothing worked.

# Description
* removes our over-strict browser-list requirements, which force a check every day

please let me know if caniuse continues to update every day.


ryanpdaley (approved):
✂️

PR #1237, opened 2024-05-16
MOVE-63

spencermountain:
# Issue Addressed
This PR closes [MOVE-63](https://move-toronto.atlassian.net/browse/MOVE-63) - sticky headers on reports 

# Description
ended up being really easy - adds a css  `position: sticky;` to report headers, so they stop scrolling, once they reach the top.
image

the grey blurred into the other rows, when they scrolled into it, so I made it blue.

multiple-table reports just work as expected. 
I tested the pdfs, and they are unchanged, and not blue.

### DIFF
--- a/web/components/FcDrawerViewCollisionReports.vue
+++ b/web/components/FcDrawerViewCollisionReports.vue
@@ -127,7 +127,7 @@
         <v-divider></v-divider>
       </div>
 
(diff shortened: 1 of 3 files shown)


ryanpdaley (approved):
🍯 (sticky!)

PR #1238, opened 2024-05-17
add ctrl-click handler

spencermountain:
# Issue Addressed
This PR closes #X [MOVE-1217](https://move-toronto.atlassian.net/browse/MOVE-1217) - add ctrl+click to study view btn.

# Description
using the neat vue `@click.ctrl` modifier, adds a 2nd callback that opens the route in a new tab, instead of the current tab.
passes an event, so it can stop propagation to the regular click handler. Seems to work 👍


ryanpdaley (approved):
Nice!

PR #1242, opened 2024-05-21
fixed the corridor error bug

lowsound42:
# Issue Addressed
This is an amendment to [MOVE-59](https://move-toronto.atlassian.net/browse/MOVE-59).

# Description
The previous change was only really tested with POINTS and not CORRIDOR type selections. The behaviour broke when a corridor was selected since we were using the location selection and not a list of all locations (which includes the spaces in between two selections). In the case of a POINTS selection, these intermediates will be dropped when generating the report but will remain in play when a CORRIDOR is selected.

### DIFF
--- a/web/components/FcDrawerViewCollisionReports.vue
+++ b/web/components/FcDrawerViewCollisionReports.vue
@@ -423,9 +423,9 @@ export default {
       this.loadingReportLayout = false;
 
       this.reportLayout = reportLayout;
-      // if (this.isDirectoryReport) {
-      //   this.extractMvcrRows(this.reportLayout);
-      // }
+      if (this.isDirectoryReport) {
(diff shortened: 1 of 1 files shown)


spencermountain (approved):

web/components/FcDrawerViewCollisionReports.vue
```diff
@@ -423,9 +423,9 @@ export default {
       this.loadingReportLayout = false;
 
       this.reportLayout = reportLayout;
-      if (this.isDirectoryReport) {
-        this.extractMvcrRows(this.reportLayout);
-      }
+      // if (this.isDirectoryReport) {
```
ryanpdaley, line 426: Should we remove this?

ryanpdaley (approved):

PR #1244, opened 2024-05-21
MOVE:1194: Moved filters to end of page, filters section removed if there are no…

imadsyed333:
# Issue Addressed
This issue closes MOVE-1194. 

# Description
Moves collision filters to the last page of pdf exports of collision directory reports. Also, the collision filters section does not appear if there are no filters.

(No filters)
image

(Filters)
image

# Tests
Tested in MOVE local v1.12.0 using Firefox.


spencermountain (approved):

ryanpdaley (approved):
📝

PR #1247, opened 2024-05-21
MOVE 59 -> downloading multiple report experiment

lowsound42:
# Issue Addressed
Again, MOVE-59.

# Description
We want to display reports separately on the web app, that's fine and is working. Downloading reports for individual points in a multi selection is great as well. What wasn't working was the downloading of reports for the corridor locations (i.e. the potential scores of locations between our actual selections).

This change takes the corridor and breaks it up into a series of points and then generates a report for each of those points. This is a test to see how long report generation takes on the dev server.

### DIFF
--- a/web/components/data/FcViewDataAggregate.vue
+++ b/web/components/data/FcViewDataAggregate.vue
@@ -135,7 +135,7 @@
 <script>
 import { mapGetters, mapMutations, mapState } from 'vuex';
 
-import { ReportExportMode } from '@/lib/Constants';
+import { ReportExportMode, LocationSelectionType } from '@/lib/Constants';
 import {
   getCollisionsByCentrelineSummary,
   getCollisionsByCentrelineSummaryPerLocation,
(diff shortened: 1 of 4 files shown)


web/components/data/FcViewDataAggregate.vue
```diff
@@ -265,12 +266,26 @@ export default {
...
     async actionDownloadReportFormatCollisions(reportFormat) {
-      const job = await postJobGenerateCollisionReports(
-        this.auth.csrf,
-        this.locationsSelection,
-        this.filterParamsCollision,
-        reportFormat,
-      );
+      let job;
+      if (this.locationsSelection.selectionType === LocationSelectionType.CORRIDOR) {
+        this.locationsSelectionForCorridorReport = JSON.parse(
```
lowsound42, line 271: I dont love this but I need a deep copy of the locationsSelection object and this was just the easiest to implement. I guess if we need to do this somewhere else, it would make sense to create a utility function that does the cloning a bit more efficiently.

spencermountain (approved):

ryanpdaley (approved):
🆒

PR #1254, opened 2024-05-23
Move 1214

lowsound42:
# Issue Addressed
Addressing MOVE-1214 and undoing issues I caused for MOVE-1189

# Description
Due to changes in move-1214 pushed earlier, I moved some logic out of the StudyRequestForm component and moved it into the individual field components. This caused an issue with the studyType input where if a study was of the ATR_SPEED_VOLUME type, on the edit page it defaulted to nothing since the logic on the parent component was removed. This makes no difference on **new** studies since they are populated by default as ATR_SVC. On editing older studies with ATR_SPEED_VOLUME we have to check if this is the case and then conditionally change them to ATR_SVC. I've put that logic back in now in a beforeMount hook.

Sorry for the confusion.


ryanpdaley (approved):
Looks good!

PR #1255, opened 2024-05-23
MOVE-1229: Map / Table interactions gone after request status change

imadsyed333:
# Issue Addressed
This PR closes MOVE-1229.

# Description
MouseEvent Listeners on table rows are now being added to the whole document body, instead of the rows themselves. This keeps the listeners "attached" to the table row, even if row data gets altered. The code for adding these listeners is now housed in `FcRequestStudyBulk.vue` instead of `FcDataTable.vue`, to ensure that the listeners only activate on the Project View page, and not on every page that contains an FcDataTable.

# Tests
Tested in MOVE local v1.12.0 using Firefox.

### DIFF
--- a/web/components/FcDataTable.vue
+++ b/web/components/FcDataTable.vue
@@ -40,7 +40,6 @@
 </template>
 
 <script>
-import { mapMutations } from 'vuex';
 import { Ripple } from 'vuetify/lib/directives';
 
 import { KeyCode } from '@/lib/Constants';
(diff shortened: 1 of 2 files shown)


spencermountain:
can you use the vue ` <tr @mouseover="myFn" @mouseleave="myExitFn">`  way?
I think those get updated and cleaned-up when the html changes automatically?

imadsyed333:
@spencermountain I wish I could. As far as I know, in newer versions of Vuetify, you can pass mouseevent listeners as props into a VDataTable, and have those listeners applied to the table rows. But I think this only works if we populate the data table with <tr>'s in the vue file....and unfortunately we use slots 😢. 

I installed vue 2.7 at some point to test this out, but I couldn't get it to work.

spencermountain (approved):

spencermountain:
ah, of course! good point.

ryanpdaley (approved):
🖱️ 👆

imadsyed333:
Found some unexpected behaviors. Closing this PR.

PR #1258, opened 2024-05-24
Rescue studies

imadsyed333:
# Issue Addressed
Study reports couldn't be downloaded as CSVs.

# Description
Study reports couldn't be downloaded as CSVs. Both single and bulk downloads were failing. The ReportController was treating these downloads as collision directory reports and attempting to add filters to them, causing havoc.

# Tests
Tested in Firefox, MOVE local v1.12.0


lib/filters/prettyCollisionFilters.js
lowsound42, line 1: Gotta rename the file as well

lowsound42:
Besides the filename thing this looks good. Should solve the issue for single and bulk study downloads.

lowsound42 (changes requested):
Just the prettyCollisionFilter file name change to be PascalCase

lib/controller/ReportController.js
```diff
@@ -153,7 +153,8 @@ ReportController.push({
 
       response.header('Content-Disposition', `attachment; filename="${key}"`);
 
-      if (singleFile && reportFormat === ReportFormat.CSV) {
+      if (singleFile && reportFormat === ReportFormat.CSV
+        && reportType === ReportType.COLLISION_DIRECTORY) {
```
lowsound42, line 157: *chefs kiss*

lowsound42 (approved):

lowsound42:
NICE

PR #1261, opened 2024-05-30
MOVE-1189   After editing the location of a request, other study values switch to defaults

ryanpdaley:
# Issue Addressed
"This PR closes [MOVE-1189](https://move-toronto.atlassian.net/browse/MOVE-1189) - After editing the location of a request, other study values switch to defaults

# Description
If the new location is 'compatible' with the current study request, all that will change is the location.
If not, a modal will appear informing the user of this:
image

# Tests
All tests are passing.

### DIFF
--- a/web/components/geo/map/EditStudyLocationPopUp.vue
+++ b/web/components/geo/map/EditStudyLocationPopUp.vue
@@ -8,8 +8,8 @@
       textOk="OK" title="Invalid location for Study Type" okButtonType="primary">
       <span class="body-1">
         We cannot conduct your study at this location.<br />
-        The Study Type will need to be saved as one that is
-        valid for this location before selecting this location.
+        Ensure that the Study Type selected for this location is valid before proceeding with the
+        selection.
       </span>
     </FcDialogAlert>
   </div>
--- a/web/components/geo/map/EditStudyLocationPopUp.vue
+++ b/web/components/geo/map/EditStudyLocationPopUp.vue
@@ -3,15 +3,15 @@
     <FcButton type="tertiary" @click="actionSelected">
       Set Location
     </FcButton>
-    <FcDialogConfirm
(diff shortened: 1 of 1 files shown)


lowsound42:
Looks good when you move from midblock to intersection (and vice versa) when the study selection is the default. i.e., if you are on a midblock with a SVC and you switch, the popup shows up and you then move to the new location with a blank study type. But if we were on a midblock on a TMC for instance, we receive the same popup telling us that the study type doesn't exist, and then we move to the new location keeping the study type as TMC. It seems like it may be a bit confusing. Would it make sense to just make the study type blank regardless of whether it exists or not so that the user reads the popup and has to make the change as suggested?

Eg: 
image
We're on a midblock on a TMC study. I want to switch to an intersection.
image
This is a bit confusing because TMC's should be valid.

image
The change is made and the studytype remains valid. 

For this user flow the popup seems to be an extra step that may add confusion and complexity? That's just my initial impression.

ryanpdaley:
@lowsound42 - hmm... I'll look into why you are seeing the popup for the TMC... you shouldn't be.

Initially I was playing around with modifying the study request too - I'm currently only updating or not updating the location - but it was like wack-a-mole. If I change the study type, then the times might not be valid, and the 'urgent' info is not contained only in the study request, and the location object in EditStudyLocationPopUp.vue isn't what the study request uses... so I figure that modal + a blank study type was the most direct way of getting a response from the user.

lowsound42:
Yeah this a tricky interaction, I did not enjoy my time with it on that new study request stuff.

spencermountain (approved):
<borat>very nice!</borat>

ryanpdaley:
image

new modal copy?

lowsound42:
> image
> 
> new modal copy?

yeah that works!

lowsound42 (approved):

PR #1267, opened 2024-06-04
MOVE-1230: UI issues when adding request to new project

imadsyed333:
# Issue Addressed
This PR closes MOVE-1230.

# Description
Forms related to study request projects now validate on submit, improving user experience. Cursor now automatically lands on first input on form load. Some inputs have new labels as well.

image

Validation on submission implemented via sending a boolean prop from the parent form to the children input fields, signalling them to display errors if needed. Timeout function is called on submission to ensure that the emails field validates its input.

# Tests
Tested in MOVE local v1.12 in Firefox, by performing the following actions: Adding to existing project, adding to new project, editing a project.

### DIFF
--- a/web/components/FcDrawerRequestStudyBulkEdit.vue
+++ b/web/components/FcDrawerRequestStudyBulkEdit.vue
@@ -63,7 +63,6 @@ import FcStudyRequestBulkDetails
 import FcNavStudyRequest from '@/web/components/requests/nav/FcNavStudyRequest.vue';
 import FcMixinLeaveGuard from '@/web/mixins/FcMixinLeaveGuard';
 import FcMixinRouteAsync from '@/web/mixins/FcMixinRouteAsync';
-import { getFirstErrorText, scrollToFirstError } from '@/web/ui/FormUtils';
 
 export default {
   name: 'FcDrawerRequestStudyBulkEdit',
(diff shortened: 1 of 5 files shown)


web/components/FcDrawerRequestStudyBulkEdit.vue
```diff
@@ -102,17 +103,15 @@ export default {
   },
   methods: {
     actionSave() {
-      if (!this.formValid) {
-        const $form = this.$refs.formWrapper;
-        scrollToFirstError($form, '.v-input');
+      setTimeout(() => {
```
spencermountain, line 106: wanna put some milliseconds on this? `setTimeout(fn, ms)`
is there a way to hook this up after the route change?

imadsyed333: If I put the timeout after the route change, then the emails field would start accepting bad values for emails. I'll give it some milliseconds though.

spencermountain: is there a way to do `.then()` after the route change? I think evan has used `Vue.nextTick()` also.
I'm just worried that your fix will only work sometimes, depending on how fast the browser/network is being.

imadsyed333: Hmm, that is true. I'm currently looking to into using refs and nextTick. Haven't gotten it to work yet though.

spencermountain: nice one!!

spencermountain (approved):

lowsound42 (approved):

PR #1273, opened 2024-06-10
MOVE-1239: Marker highlighting bug on track request project page

imadsyed333:
# Issue Addressed
This PR closes MOVE-1239.

# Description
"Bad" variable types being passed into functions caused unexpected behaviour. Added logic to check for variable types, and ensure they are "good". 

Added new computed property to help identify different study requests that share the same location marker. When hovering between two study requests that share the same location marker, the location marker will remain selected.

# Tests
Tested in MOVE local v1.13 using Firefox

### DIFF
--- a/web/components/geo/map/FcMap.vue
+++ b/web/components/geo/map/FcMap.vue
@@ -366,13 +366,14 @@ export default {
       }
       return !this.drawerOpen || !featureMatchesRoute;
     },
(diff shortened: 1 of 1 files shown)


spencermountain (approved):
nice one

ryanpdaley (approved):

Issue #981, opened 2021-06-24
Use ST_AsMVT directly in DynamicTileDAO


candu:
**Description**
With #955 , MOVE's database will be upgraded to PostgreSQL 11.x, which [includes support for `libprotobuf`](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.version11).  This in turn enables use of the PostGIS function [`ST_AsMVT`](https://postgis.net/docs/ST_AsMVT.html), which allows us to generate vector tiles directly in database.

This task refactors `DynamicTileDAO` to use `ST_AsMVT`.

**Acceptance Criteria**
- [ ] replace current process (see Additional Notes) with `ST_AsMVT`;
- [ ] ensure that tiles load successfully in frontend.

**Additional Notes**
Our current setup has three steps:

1. fetch features in tile coordinates using previous-generation PostGIS functions, in `DynamicTileDAO`;
2. in `VectorTile`, convert those features to a representation that can be used with `vt-pbf`;
3. in `DynamicTileController`, use `vt-pbf` to make the final conversion to `.pbf` vector tiles.

It's likely (though not certain) that switching to `ST_AsMVT` will also improve performance.

candu:
Adding @leonL - this is a good first issue that focuses on the backend / database parts of MOVE, specifically `DynamicTileDAO`.  In this task, you'll change the queries in that file to use [`ST_AsMVT`](https://postgis.net/docs/ST_AsMVT.html), a function in newer versions of PostGIS that allows you to generate vector tiles _directly within PostgreSQL_.

leonL:
Migrated to jira [MOVE-209]

PR #1279, opened 2024-06-14
MOVE-209: Use ST_AsMVT directly in DynamicTileDAO

imadsyed333:
# Issue Addressed
This PR closes MOVE-209.

# Description
Edited queries in `DynamicTileDAO.js` to take advantage of the `ST_AsMVT` function in Postgis. This allows the database to serve vector tiles to the app, easing the load on the backend.

# Tests
Tested in MOVE local v1.13 using Firefox


ryanpdaley (approved):
I think this looks good, and seems to match the documentation on:

- https://github.com/CityofToronto/bdit_flashcrow/issues/981
- https://postgis.net/docs/ST_AsMVT.html

I guess we'll keep an eye on it to see if this causes db load, and if it makes the tiles snappier from the backend! 👍

PR #1288, opened 2024-07-12
Update Backend for ADFS

ryanpdaley:
# Issue Addressed
This PR closes [MOVE-1263](https://move-toronto.atlassian.net/browse/MOVE-1263) and [MOVE-1265](https://move-toronto.atlassian.net/browse/MOVE-1265)
image

# Description
Adds the extra fields that we can currently get from ADFS to the database: 

- `userTitle`: The title of the user within the City of Toronto
- `firstName`: The user's first name
- `lastName`: The user's last name
- `department`: The user's department within the City of Toronto

I'm not sure which of these will be populated within ADFS for production accounts, so they will be returned as `null` if they aren't defined or are blank. 

There is also an issue that we were using the `sub` value as the unique key for a user and in local the new `client_id` is returning new values for this, so get duplicate values. I've tried to mitigate this by looking for an email match, and updating the user record (including the `sub` value) if a `sub` match isn't found.

# Tests
All existing tests are passing, but no new tests were created for this yet.


spencermountain:
TIL nodejs does not have a atob method. LGTM!

spencermountain (approved):

lowsound42 (approved):
a brave new world of user identity details

mkewins:
cool!!

PR #1290, opened 2024-07-18
MOVE-1251: Add data source to collision filters

imadsyed333:
# Issue Addressed
This PR addresses MOVE-1251.

# Description
Added data source as a collision filter. The sources are presented as checkboxes in the filter drawer.
image
This filter will filter collisions just like any other filter.

# Tests
Tested in MOVE local v1.13 in Firefox and Chrome


lowsound42 (approved):

ryanpdaley (approved):
🚀

lib/db/filters/CollisionFiltersSql.js
```diff
@@ -50,6 +51,11 @@ function getCollisionFilters(features, collisionQuery) {
       .join(' OR ');
     filters.push(`(${filterDetails})`);
   }
+  if (sources.length > 0) {
+    params.sources = sources;
+    const filterSources = sources.map(({ field }) => `'${field}'`).join(', ');
```
spencermountain, line 56: I assume there's no stuff inside the value here, like a `'` or `,` that would bung-up the sql query.

spencermountain:
this is so cool Imad!

